PR #8, opened 2016-10-08
Switch from pandoc to Jekyll

SakalisC:
Pandoc is not easy to use because the constant changes in its output style force us to use a very specific pandoc version. Jekyll on the other hand not only allows us to ditch the html files and use just markdown, but it is also the recommended way for publishing markdown sites on Github.

Preview available on https://sakalisc.github.io/argodsm/


SakalisC:
Jekyll + Github guide: https://help.github.com/articles/setting-up-your-github-pages-site-locally-with-jekyll/

SakalisC:
I have updated the commit to use Jekyll's `{% include %}` for the two example C++ files. There is a known issue that introduces an extra newline at the end of the included content but there isn't an easy solution for it.

PR #10, opened 2016-10-26
ensure MPI progress on release synchronization

davidklaftenegger:
As MPI attempts to delay communication as long as possible
our internal synchronization needs to poke the MPI subsystem
even when no data is read from the network.


davidklaftenegger:
@magnusandersnorgren @SakalisC @lundgren87 @pekemark:
The "questionable thing" mentioned by Magnus is still unanswered. Does anyone have any insight on this?

PR #12, opened 2016-11-03
fix access to private remote pages

davidklaftenegger:
When accessing a private page on a remote homenode, it could happen that
an undefined node would be accessed. As the node is in private state, no
access is actually needed.

### DIFF
--- a/src/backend/mpi/swdsm.cpp
+++ b/src/backend/mpi/swdsm.cpp
@@ -770,7 +770,7 @@ void * loadcacheline(void * x){
 		unsigned long offset = getOffset(lineAddr);
 		if(isPowerOf2((tempsharer)&invid) && tempsharer != id && prevsharer == 0){ //Other private. but may not have loaded page yet.
 			unsigned long ownid = tempsharer&invid; // remove own bit
-			static const unsigned long invalid_owner = static_cast<unsigned long>(-1);
+			constexpr unsigned long invalid_owner = static_cast<unsigned long>(-1);
 			unsigned long owner = invalid_owner; // initialize to failsafe value
 			for(n=0; n<numtasks; n++) {
 				if(1ul<<n==ownid) {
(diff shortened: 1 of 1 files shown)


src/backend/mpi/swdsm.cpp
```diff
@@ -769,17 +769,20 @@ void * loadcacheline(void * x){
 
 		unsigned long offset = getOffset(lineAddr);
 		if(isPowerOf2((tempsharer)&invid) && tempsharer != id && prevsharer == 0){ //Other private. but may not have loaded page yet.
-			long ownid = (long)(tempsharer&invid);
-			unsigned long owner;
-			for(n=0; n<numtasks; n++){
-				if((unsigned)(1<<n)==(unsigned)ownid){
+			unsigned long ownid = tempsharer&invid; // remove own bit
+			static const unsigned long invalid_owner = static_cast<unsigned long>(-1);
```
SakalisC: Can this be in some larger scope to avoid having to create it twice? Also, the static is fairly useless, if not harmful, since any semi-decent compiler will constant propagate the value.

I would suggest using a constexpr in global scope: `constexpr unsigned long invalid_owner = static_cast<...>(-1);`

I assume that the static cast is used to silence compiler warnings.

davidklaftenegger: you are correct about the static const, should be constexpr.
static_cast is used to document that this is an invalid value for the type in question, I have not checked whether any or all compilers accept this silently.

However, I am not sure that creating this at the global scope is helpful or harmful to readability of the code. If used at global code it definitely should be static to avoid making it part of the ABI, right?

SakalisC: Readability shouldn't be an issue as the actual value is not important, it's just some value that signifies an invalid node ID. Just like any global constant or #define value.

I agree with making it static, but using namespace {} instead of static.

davidklaftenegger: resolved in 03424b5 and 3e1353d

PR #13, opened 2016-11-15
add signal handler module

davidklaftenegger:
Segmentation faults need to be handled explicitly by ArgoDSM.
However, handling them directly in the backend causes problems
with applications that also want to handle signals. This module
allows to "multiplex" signals to the correct handler.

### DIFF
--- a/src/backend/singlenode/singlenode.cpp
+++ b/src/backend/singlenode/singlenode.cpp
@@ -5,6 +5,7 @@
  */
 
 #include "data_distribution/data_distribution.hpp"
+#include "signal/signal.hpp"
 #include "synchronization/global_tas_lock.hpp"
 #include "types/types.hpp"
 #include "virtual_memory/virtual_memory.hpp"
@@ -12,11 +13,13 @@
 
 #include <atomic>
 #include <condition_variable>
+#include <csignal>
 #include <cstddef>
 #include <cstring>
 #include <mutex>
 
 namespace vm = argo::virtual_memory;
+namespace sig = argo::signal;
 
 /** @brief a lock for atomically executed operations */
 std::mutex atomic_op_mutex;
@@ -54,6 +57,19 @@ char* memory;
  */
 std::size_t memory_size;
 
+/**
+ * @brief a dummy signal handler function
+ * @warning this function is not strictly portable because it resets the handler and re-raises the signal
+ * @param sig signal number
+ * @see check `man sigaction` for additional information
+ */
+void singlenode_handler(int sig, siginfo_t*, void*) {
+	printf("A segfault was encountered in ArgoDSM memory. "
+		"This should never happen, as the singlenode backend is in use.\n"
+	);
+	std::signal(sig, SIG_DFL);
+	std::raise(sig);
+}
 namespace argo {
 	namespace backend {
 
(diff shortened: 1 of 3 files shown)


src/signal/signal.hpp
```diff
@@ -0,0 +1,106 @@
...
+				 * @param si additional signal information
+				 * @param unused traditionally unused parameter
+				 * @see check `man sigaction` for additional information
+				 */
+				static void argo_signal_handler(int sig, siginfo_t *si, void *unused) {
+					namespace vm = argo::virtual_memory;
+					const auto addr = si->si_addr;
+					const auto start = static_cast<char*>(vm::start_address());
+					const auto end = start + vm::size();
+					if (addr < start || addr > end) {
```
davidklaftenegger: yes

src/backend/singlenode/singlenode.cpp
```diff
@@ -54,6 +57,21 @@ char* memory;
...
+ * @warning this function is not strictly portable because it resets the handler and re-raises the signal
+ * @param sig signal number
+ * @param si additional information
+ * @param unused traditionally unused
+ * @see check `man sigaction` for additional information
+ */
+void handler(int sig, siginfo_t* si, void* unused) {
+	(void) si;
+	(void) unused;
+	printf("A segfault was encountered in ArgoDSM memory. This should never happen.");
```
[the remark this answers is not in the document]
davidklaftenegger: Do you mean a comment in the code, or should the printf be more specific?
Also, should the printf include a newline?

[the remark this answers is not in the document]
SakalisC: > Do you mean a comment in the code, or should the printf be more specific?

I would go for the latter, but either is fine with me.

>  Also, should the printf include a newline?

I guess so.

src/backend/singlenode/singlenode.cpp
```diff
@@ -54,6 +57,21 @@ char* memory;
...
 
+/**
+ * @brief a dummy signal handler function
+ * @warning this function is not strictly portable because it resets the handler and re-raises the signal
+ * @param sig signal number
+ * @param si additional information
+ * @param unused traditionally unused
+ * @see check `man sigaction` for additional information
+ */
+void handler(int sig, siginfo_t* si, void* unused) {
```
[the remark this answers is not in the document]
davidklaftenegger: This works with doxygen as well; but reveals that it gets confused by two functions of the same name in different backends. Fixing it.

src/backend/singlenode/singlenode.cpp
```diff
@@ -62,6 +78,7 @@ namespace argo {
 			memory_size = size;
 			using namespace data_distribution;
 			naive_data_distribution<0>::set_memory_space(nodes, memory, size);
+			sig::signal_handler<SIGSEGV>::install_argo_handler(&singlenode_handler);
```
SakalisC: Should a backend that doesn't use a signal handler install one in the first place? Isn't this just extra overhead?

davidklaftenegger: In principle you are right, this serves no "purpose". However, this is true of the entire singlenode backend, which you yourself suggested to be renamed to "dummy". For the purpose of a dummy, I believe the intention of this code is to test that the API is properly defined and teach how to apply it. Therefore I would argue against removing the code.

SakalisC: OK, makes sense. This wasn't a blocking comment anyway.

PR #14, opened 2016-11-18
Cache backend fixes

magnusandersnorgren:
Removes the unnecessary memsets in the MPI-backend and makes it possible to start a backend with a specified amount of cache.  (Tested on 2 nodes on tintin)

### DIFF
--- a/src/backend/backend.hpp
+++ b/src/backend/backend.hpp
@@ -41,13 +41,14 @@ namespace argo {
 	namespace backend {
 		/**
 		 * @brief initialize backend
-		 * @param size the size of the global memory to initialize
+		 * @param argo_size the size of the global memory to initialize
+		 * @param cache_size the size of the cache used by ArgoDSM in number of bytes, may be unused in some backends
 		 * @warning the signature of this function may change
 		 * @todo maybe this should be tied better to the concrete memory
 		 *       allocated rather than a generic "initialize this size"
 		 *       functionality.
 		 */
-		void init(std::size_t size);
+		void init(std::size_t argo_size, std::size_t cache_size);
 
 		/**
 		 * @brief get ArgoDSM node ID
--- a/src/backend/mpi/swdsm.cpp
+++ b/src/backend/mpi/swdsm.cpp
@@ -1044,15 +1044,15 @@ void set_sighandler(){
 	sigaction(SIGSEGV, &sa, &old_sigaction);
 }
 
-void argo_initialize(unsigned long long size){
+void argo_initialize(unsigned long long argo_size, size_t cache_size){
 	int i;
 	unsigned long j;
 	initmpi();
 	unsigned long alignment = pagesize*CACHELINE*numtasks;
-	if((size%alignment)>0){
-		size += alignment - 1;
-		size /= alignment;
-		size *= alignment;
+	if((argo_size%alignment)>0){
+		argo_size += alignment - 1;
+		argo_size /= alignment;
+		argo_size *= alignment;
 	}
 
 	startAddr = vm::start_address();
@@ -1065,7 +1065,7 @@ void argo_initialize(unsigned long long size){
 		pthread_barrier_init(&threadbarrier[i],NULL,i);
 	}
 
-	cachesize = size+pagesize*CACHELINE;
+	cachesize = cache_size+pagesize*CACHELINE;
 	cachesize /= pagesize;
 	cachesize /= CACHELINE;
 	cachesize *= CACHELINE;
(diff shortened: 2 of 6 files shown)


src/backend/mpi/swdsm.cpp
```diff
@@ -1044,15 +1044,15 @@ void set_sighandler(){
 	sigaction(SIGSEGV, &sa, &old_sigaction);
 }
 
-void argo_initialize(unsigned long long size){
+void argo_initialize(unsigned long long argo_size, size_t cache_size){
```
davidklaftenegger: Should we use this opportunity to change this from unsigned long long to size_t?

src/backend/backend.hpp
```diff
@@ -41,13 +41,14 @@ namespace argo {
 	namespace backend {
 		/**
 		 * @brief initialize backend
-		 * @param size the size of the global memory to initialize
+		 * @param argo_size the size of the global memory to initialize
```
davidklaftenegger: this should also specify the unit of the memory

src/backend/backend.hpp
```diff
@@ -41,13 +41,14 @@ namespace argo {
...
+		 * @param cache_size the size of the cache used by ArgoDSM in number of bytes, may be unused in some backends
 		 * @warning the signature of this function may change
 		 * @todo maybe this should be tied better to the concrete memory
 		 *       allocated rather than a generic "initialize this size"
 		 *       functionality.
 		 */
-		void init(std::size_t size);
+		void init(std::size_t argo_size, std::size_t cache_size);
```
SakalisC: Should `cache_size` have a default value?

davidklaftenegger: Would that be for backwards compatibility, or why should the backend have a default value for this?

src/mempools/global_mempool.hpp
```diff
@@ -50,7 +50,7 @@ namespace argo {
 				 */
 				global_memory_pool(std::size_t size) {
 					size+=reserved;
-					backend::init(size);
+					backend::init(size, 1<<30); // 1GB of memory for cache as default (= 1<<30 bytes)
```
davidklaftenegger: Do we want to make this changeable without recompiling?

src/backend/backend.hpp
SakalisC: I don't remember any more, but I would make more sense to have the default argument in the frontend part.

davidklaftenegger: well, the frontend code hard-codes the value to 1G..

src/backend/singlenode/singlenode.cpp
```diff
@@ -57,7 +57,7 @@ std::size_t memory_size;
 namespace argo {
 	namespace backend {
 
-		void init(std::size_t size) {
+		void init(std::size_t argo_size, std::size_t cache_size){
```
davidklaftenegger: renamed parameters should also be renamed in the function body, currently this does not even compile

src/mempools/global_mempool.hpp
davidklaftenegger: The default value of 1GB already slows down the tests significantly.
Using a larger value, like 8GB causes frequent test timeouts even on a single node.
This may or may not be linked to the fact that the tests do not allocate that much memory, thus the cache would be larger than the distributed memory area.

src/backend/mpi/swdsm.cpp
```diff
@@ -1199,14 +1199,7 @@ void argo_initialize(unsigned long long size){
...
 
-	memset(pagecopy, 0, cachesize*pagesize);
-	memset(touchedcache, 0, cachesize);
-	memset(globalData, 0, size_of_chunk*sizeof(argo_byte));
-	memset(cacheData, 0, cachesize*pagesize);
-	memset(lockbuffer, 0, pagesize);
-	memset(globalSharers, 0, gwritersize);
-	memset(cacheControl, 0, cachesize*sizeof(control_data));
-
+	//Sets cache control structures to initial state.
```
davidklaftenegger: My understanding is that this part (removing initialization) has nothing to do with the remaining patch.
I am not sure about the performance impacts of this part yet, but maybe we should split this off into a separate pull-request?

src/backend/mpi/swdsm.cpp
davidklaftenegger: addressed in d762963

src/backend/backend.hpp
davidklaftenegger: addressed in d762963

src/mempools/global_mempool.hpp
davidklaftenegger: addressed in d762963

src/backend/singlenode/singlenode.cpp
davidklaftenegger: addressed in 87ef1b1

src/mempools/global_mempool.hpp
davidklaftenegger: fixed in 1a3dc8a86acbf21bdbc3650c8230326a4d87ca7d and #38

PR #18, opened 2019-05-14
Coherence fix

davidklaftenegger:
The current use of MPI_LOCK_SHARED on updates to the cache metadata allows for a race condition when multiple nodes are attempting to declare themselves writers of an entry at the same time. This can lead to lost updates, where some node does not know about other nodes also writing to a page, ultimately leading to loss of coherence.

This patch contains two commits:
In the first, MPI_LOCK_EXCLUSIVE is applied to read-modify-write operations, which should avoid the lost update issue.
The second commit adds MPI_LOCK_EXCLUSIVE to another read-only instance of the same window.

### DIFF
--- a/src/backend/mpi/swdsm.cpp
+++ b/src/backend/mpi/swdsm.cpp
@@ -368,7 +368,7 @@ void handler(int sig, siginfo_t *si, void *unused){
 		unsigned long prevsharer = (globalSharers[classidx])&id;
 		MPI_Win_unlock(workrank, sharerWindow);
 		if(prevsharer != id){
-			MPI_Win_lock(MPI_LOCK_SHARED, workrank, 0, sharerWindow);
+			MPI_Win_lock(MPI_LOCK_EXCLUSIVE, workrank, 0, sharerWindow);
 			sharers = globalSharers[classidx];
 			globalSharers[classidx] |= id;
 			MPI_Win_unlock(workrank, sharerWindow);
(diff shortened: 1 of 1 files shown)


SakalisC:
1. In [line 390 of swdsm.cpp file](https://github.com/etascale/argodsm/blob/7c1d5b1738f1cbe7d0b5bacd03cf1a712420ec5a/src/backend/mpi/swdsm.cpp#L390) there is an accumulate operation with a comment claiming that it updates the remove private holder to shared. It uses a SHARED lock again, can this cause an issue? A similar pattern can be seen in other places, this is just the first one I saw.

2. I noticed while going thought the code that the `self_invalidation` function was accessing the `globalSharers` without a lock myself, so I agree with this change.

3. The places where the locks were changed from SHARED to EXCLUSIVE make sense.

4. I see that the `globalSharers` is [still mapped into the Argo memory](https://github.com/etascale/argodsm/blob/7c1d5b1738f1cbe7d0b5bacd03cf1a712420ec5a/src/backend/mpi/swdsm.cpp#L1172)? Is this because otherwise it will not be mapped for RDMA operations? Is it possible to map it for RDMA operations without mapping it in the Argo memory space?

SakalisC:
To clarify (4) a little bit, how is `globalSharers` affected by being mapped in the Argo memory space? I am confused by it.

davidklaftenegger:
Thank you for your points, you are pointing out a real issue.

point 1) I amended the commit to also put MPI_LOCK_EXCLUSIVE around MPI_Accumulate calls, as these also write to shared locations.

points 2 and 3 require no further changes.

point 4) You are correct that it is still mapped into ArgoDSM's global space. I currently think this is done to align the addresses for the metadata across all nodes. This is not strictly required, and doing RDMA on the metadata should be possible without aligned addresses as well, but I think changing this in the code is better left to a future patch.

SakalisC:
For (1), it looks like all the relevant `MPI_Accumulate` calls are now protected by EXCLUSIVE locks.
The `MPI_Get_accumulate` calls that are left all seem to write to a local variable, so SHARED locks should be fine.

For (4), as long as the coherence of Argo does not interact with the RDMA operations in some weird way, it's fine with me.

👍

PR #23, opened 2019-10-02
Limit cache size

davidklaftenegger:
This PR supersedes #19, which I accidentally broke.

This patch contains several commits:

d5b1670 gives the responsibility of initialising everything to argo::init()
32cbb09  adds the option to use an environment variable to set the memory size used
3187658 only renames arguments from size to argo_size
f08ae31 limits the cache size to a value also optionally set through argo::init() or an environment variable
32bbf41 limits the cache size to the total memory size, if it is smaller than the requested memory size

The main changes in 3187658 and f08ae31 are taken from #14, but split into two commits for readability.
Note that this pull request does not contain all features from #14, as it only deals with commit
60e68ce.

### DIFF
--- a/src/argo.h
+++ b/src/argo.h
@@ -14,11 +14,11 @@
 
 /**
  * @brief initialize ArgoDSM system
- * @param size The desired size of the global memory in bytes, or 0. If the
- *             value is specified as 0, then the value in environment
- *             variable @ref ARGO_MEMORY_SIZE is used instead.
+ * @param argo_size The desired size of the global memory in bytes, or 0. If the
+ *                  value is specified as 0, then the value in environment
+ *                  variable @ref ARGO_MEMORY_SIZE is used instead.
  */
-void argo_init(size_t size);
+void argo_init(size_t argo_size);
 
 /**
  * @brief shut down ArgoDSM system
--- a/src/argo.h
+++ b/src/argo.h
@@ -14,7 +14,9 @@
 
 /**
  * @brief initialize ArgoDSM system
- * @param size the desired size of the global memory in bytes
+ * @param size The desired size of the global memory in bytes, or 0. If the
+ *             value is specified as 0, then the value in environment
+ *             variable @ref ARGO_MEMORY_SIZE is used instead.
  */
 void argo_init(size_t size);
 
--- a/src/argo.h
+++ b/src/argo.h
@@ -17,9 +17,11 @@
  * @param argo_size The desired size of the global memory in bytes, or 0. If the
  *                  value is specified as 0, then the value in environment
  *                  variable @ref ARGO_MEMORY_SIZE is used instead.
+ *                  If @ref ARGO_MEMORY_SIZE is unset, then a default is used.
  * @param cache_size The desired size of the local cache in bytes, or 0. If the
  *                   value is specified as 0, then the value in environment
  *                   variable @ref ARGO_CACHE_SIZE is used instead.
+ *                   If @ref ARGO_CACHE_SIZE is unset, then a default is used.
  */
 void argo_init(size_t argo_size, size_t cache_size);
 
--- a/src/argo.h
+++ b/src/argo.h
@@ -17,8 +17,11 @@
  * @param argo_size The desired size of the global memory in bytes, or 0. If the
  *                  value is specified as 0, then the value in environment
  *                  variable @ref ARGO_MEMORY_SIZE is used instead.
+ * @param cache_size The desired size of the local cache in bytes, or 0. If the
+ *                   value is specified as 0, then the value in environment
+ *                   variable @ref ARGO_CACHE_SIZE is used instead.
  */
-void argo_init(size_t argo_size);
+void argo_init(size_t argo_size, size_t cache_size);
 
 /**
  * @brief shut down ArgoDSM system
--- a/src/argo.hpp
+++ b/src/argo.hpp
@@ -42,11 +42,11 @@ namespace argo {
 
 	/**
 	 * @brief initialize ArgoDSM system
-	 * @param size The desired size of the global memory in bytes, or 0.
-	 *             If the value is omitted (or specified as 0), then the value in
-	 *             environment variable @ref ARGO_MEMORY_SIZE is used instead.
+	 * @param argo_size The desired size of the global memory in bytes, or 0. If the
+	 *                  value is omitted (or specified as 0), then the value in
+	 *                  environment variable @ref ARGO_MEMORY_SIZE is used instead.
 	 */
-	void init(std::size_t size = 0);
+	void init(std::size_t argo_size = 0);
 
 	/**
 	 * @brief shut down ArgoDSM system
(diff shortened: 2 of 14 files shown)


src/backend/singlenode/singlenode.cpp
```diff
@@ -73,12 +73,11 @@ void singlenode_handler(int sig, siginfo_t*, void*) {
...
 
-		void init(std::size_t size) {
-			memory = static_cast<char*>(vm::allocate_mappable(4096, size));
-			memory_size = size;
+		void init(std::size_t argo_size) {
+			memory = static_cast<char*>(vm::allocate_mappable(4096, argo_size));
+			memory_size = argo_size;
 			using namespace data_distribution;
-			naive_data_distribution<0>::set_memory_space(nodes, memory, size);
-			sig::signal_handler<SIGSEGV>::install_argo_handler(&singlenode_handler);
```
pekemark: Is the signal handler for some reason no longer needed for singlenode? Even if the removal of the installer is correct, should it be part of this PR?

davidklaftenegger: This looks like a mistake indeed. Good catch, will be fixed later today.

src/argo.h
```diff
@@ -14,9 +14,14 @@
 
 /**
  * @brief initialize ArgoDSM system
- * @param size the desired size of the global memory in bytes
+ * @param argo_size The desired size of the global memory in bytes, or 0. If the
+ *                  value is specified as 0, then the value in environment
+ *                  variable @ref ARGO_MEMORY_SIZE is used instead.
+ * @param cache_size The desired size of the local cache in bytes, or 0. If the
+ *                   value is specified as 0, then the value in environment
+ *                   variable @ref ARGO_CACHE_SIZE is used instead.
```
SakalisC: If `ARGO_CACHE_SIZE` is uninitialized, then a fallback value is used. Should this be documented here?

src/argo.hpp
```diff
@@ -40,9 +42,14 @@ namespace argo {
 
 	/**
 	 * @brief initialize ArgoDSM system
-	 * @param size the desired size of the global memory in bytes
+	 * @param argo_size The desired size of the global memory in bytes, or 0. If the
+	 *                  value is omitted (or specified as 0), then the value in
+	 *                  environment variable @ref ARGO_MEMORY_SIZE is used instead.
+	 * @param cache_size The desired size of the local cache in bytes, or 0. If the
+	 *                   value is omitted (or specified as 0), then the value in
+	 *                   environment variable @ref ARGO_CACHE_SIZE is used instead.
```
SakalisC: Same as for the .h file.

src/backend/mpi/swdsm.cpp
```diff
@@ -1010,15 +1010,15 @@ void * argo_gmalloc(unsigned long size){
...
 	initmpi();
 	unsigned long alignment = pagesize*CACHELINE*numtasks;
-	if((size%alignment)>0){
-		size += alignment - 1;
-		size /= alignment;
-		size *= alignment;
+	if((argo_size%alignment)>0){
+		argo_size += alignment - 1;
+		argo_size /= alignment;
+		argo_size *= alignment;
```
pekemark: For a future patch: Can this alignment rounding be placed in a function? With a good function name it would probably be clearer what kind of rounding it is. There are similar occurrences elsewhere.

src/backend/mpi/swdsm.cpp
```diff
@@ -1031,7 +1031,13 @@ void argo_initialize(unsigned long long size){
...
+	cachesize = 0;
+	if(cache_size > argo_size) {
+		cachesize += argo_size;
+	} else {
+		cachesize += cache_size;
+	}
+	cachesize += pagesize*CACHELINE;
 	cachesize /= pagesize;
 	cachesize /= CACHELINE;
 	cachesize *= CACHELINE;
```
pekemark: For a future patch: There are two things going on here: First, rounding up to a whole number of cache blocks (and if the size already is, an extra cache block, which perhaps is not necessary). Second, division by `pagesize` to express `cachesize` in number of pages. These two actions should be separated for better clarity and the rounding performed by a function.

src/backend/mpi/swdsm.cpp
```diff
@@ -1073,19 +1079,19 @@ void argo_initialize(unsigned long long size){
 	MPI_Comm_create(MPI_COMM_WORLD,workgroup,&workcomm);
 	MPI_Group_rank(workgroup,&workrank);
 
-	if(size < pagesize*numtasks){
-		size = pagesize*numtasks;
+	if(argo_size < pagesize*numtasks){
+		argo_size = pagesize*numtasks;
 	}
```
pekemark: For future patch: This could be replaced with `std::max`.

src/backend/mpi/swdsm.cpp
```diff
@@ -1073,19 +1079,19 @@ void argo_initialize(unsigned long long size){
...
 
 	alignment = CACHELINE*pagesize;
-	if(size % (alignment*numtasks) != 0){
-		size = alignment*numtasks * (1+(size)/(alignment*numtasks));
+	if(argo_size % (alignment*numtasks) != 0){
+		argo_size = alignment*numtasks * (1+(argo_size)/(alignment*numtasks));
 	}
```
pekemark: For future patch: This is also rounding, should be performed by a function (see previous comments).

src/backend/singlenode/singlenode.cpp
```diff
@@ -73,12 +73,14 @@ void singlenode_handler(int sig, siginfo_t*, void*) {
...
-			memory_size = size;
+		void init(std::size_t argo_size, std::size_t cache_size){
+			/** @todo the cache_size parameter is not needed
+			 *        and should not be part of the backend interface */
+			(void)(cache_size);
+			memory = static_cast<char*>(vm::allocate_mappable(4096, argo_size));
+			memory_size = argo_size;
 			using namespace data_distribution;
-			naive_data_distribution<0>::set_memory_space(nodes, memory, size);
-			sig::signal_handler<SIGSEGV>::install_argo_handler(&singlenode_handler);
```
pekemark: Is the signal handler for some reason no longer needed for singlenode? Even if the removal of the installer is correct, should it be part of this PR? (Duplicate comment for the review.)

davidklaftenegger: Fixed in 4a3f478

src/argo.h
davidklaftenegger: Fixed in 7d2d3c7

src/argo.hpp
davidklaftenegger: Fixed in 7d2d3c7

src/backend/singlenode/singlenode.cpp
davidklaftenegger: Fixed in 4a3f478

src/backend/mpi/swdsm.cpp
davidklaftenegger: See issue #24

src/backend/mpi/swdsm.cpp
davidklaftenegger: See issue #24

src/backend/mpi/swdsm.cpp
davidklaftenegger: See issue #24

src/backend/mpi/swdsm.cpp
davidklaftenegger: See issue #24

PR #26, opened 2019-10-17
Remove helper threads

davidklaftenegger:
This pull requests removes the helper threads for writebuffer writeback, fetching, and prefetching cache entries.

The aim is to reduce complexity of the code base, as these helper threads appear to serve no purpose (other than some POSIX requirements, according to  @magnusandersnorgren, which I believe we likely break at other occasions anyway).

Despite the amount of changes, this pull request aims to not change functionality.
Some previously concurrent events (fetch & prefetch) no longer race and thus code dealing with their ordering has been removed.

### DIFF
--- a/src/backend/mpi/swdsm.cpp
+++ b/src/backend/mpi/swdsm.cpp
@@ -100,14 +100,13 @@ sem_t ibsem;
  */
 void load_cache_entry(unsigned long tag, unsigned long line);
 
-/** @brief Tracking address of pages that should be loaded by loadthread2 */
-unsigned long *prefetchline;
-/** @brief Tracking cacheindex of pages that should be loaded by loadthread2 */
-unsigned long *prefetchtag;
-/** @brief loadthread2 waits on this to start loading remote pages */
-sem_t prefetchstartsem;
-/** @brief signalhandler waits on this to complete a transfer */
-sem_t prefetchwaitsem;
+/**
+ * @brief prefetch into cache helper function, which duplicates a lot of code
+ * @param tag aligned address to prefetch into the cache
+ * @param line cache entry index to use
+ * @todo this function duplicates a lot of code from load_cache_entry(), this should be fixed
+ */
+void prefetch_cache_entry(unsigned long tag, unsigned long line);
 
 /*Global lock*/
 /** @brief  Local flags we spin on for the global lock*/
@@ -322,8 +321,6 @@ void handler(int sig, siginfo_t *si, void *unused){
 
 	unsigned long * localAlignedAddr = (unsigned long *)((char*)startAddr + lineAddr);
 	unsigned long startIndex = getCacheIndex(lineAddr);
-	unsigned long cacheIndex = getCacheIndex(alignedDistrAddr);
-	cacheIndex = startIndex;
 	alignedDistrAddr = lineAddr;
 
 	unsigned long homenode = getHomenode(lineAddr);
@@ -415,50 +412,14 @@ void handler(int sig, siginfo_t *si, void *unused){
 	tag = cacheControl[startIndex].tag;
 
 	if(state == INVALID || (tag != lineAddr && tag != GLOBAL_NULL)){
-		if(prefetchline[0] != GLOBAL_NULL){
-			if(prefetchline[0] <= startIndex && startIndex < prefetchline[0]+CACHELINE){
-				load_cache_entry((lineAddr+pagesize*CACHELINE), ((startIndex+CACHELINE)%cachesize));
-			}
-			else{
-				load_cache_entry(alignedDistrAddr, (cacheIndex%cachesize));
-			}
-
-			sem_wait(&prefetchwaitsem);
-			prefetchline[0] = GLOBAL_NULL;
-			prefetchtag[0]= GLOBAL_NULL;
-			pthread_mutex_unlock(&cachemutex);
-			double t2 = MPI_Wtime();
-			stats.loadtime+=t2-t1;
-			return;
-		}
-		else{
-			load_cache_entry(alignedDistrAddr, (startIndex%cachesize));
-
+		load_cache_entry(alignedDistrAddr, (startIndex%cachesize));
 #ifdef DUAL_LOAD
-			prefetchline[0]=(startIndex+CACHELINE)%cachesize;
-			prefetchtag[0]=alignedDistrAddr+CACHELINE*pagesize;
-			sem_post(&prefetchstartsem);
-#else
-			prefetchline[0]=GLOBAL_NULL;
-			prefetchtag[0]=GLOBAL_NULL;
-
+		prefetch_cache_entry((alignedDistrAddr+CACHELINE*pagesize), ((startIndex+CACHELINE)%cachesize));
 #endif
-			pthread_mutex_unlock(&cachemutex);
-			double t2 = MPI_Wtime();
-			stats.loadtime+=t2-t1;
-			return;
-		}
-	}
-
-	if(prefetchline[0] != GLOBAL_NULL){
-		sem_wait(&prefetchwaitsem);
-		prefetchline[0] = GLOBAL_NULL;
-		prefetchtag[0]=GLOBAL_NULL;
 		pthread_mutex_unlock(&cachemutex);
 		double t2 = MPI_Wtime();
 		stats.loadtime+=t2-t1;
 		return;
-
 	}
 
 	unsigned long line = startIndex / CACHELINE;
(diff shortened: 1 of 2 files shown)


src/backend/mpi/swdsm.cpp
```diff
@@ -95,22 +93,20 @@ char * barwindowsused;
...
-/** @brief loadthread2 waits on this to start loading remote pages */
-sem_t prefetchstartsem;
-/** @brief signalhandler waits on this to complete a transfer */
-sem_t prefetchwaitsem;
+/**
+ * @brief load into cache helper function
+ * @param tag aligned address to load into the cache
+ * @param line cache entry index to use
+ */
+void load_cache_entry(unsigned long tag, unsigned long line);
```
SakalisC: I'd rather the parameters where named something better. It seems the names where taken from the existing code? How about something like `address` and `local_index`?

src/backend/mpi/swdsm.cpp
```diff
@@ -95,22 +93,20 @@ char * barwindowsused;
...
+
+/**
+ * @brief prefetch into cache helper function, which duplicates a lot of code
+ * @param tag aligned address to prefetch into the cache
+ * @param line cache entry index to use
+ * @todo this function duplicates a lot of code from load_cache_entry(), this should be fixed
```
SakalisC: I would have preferred for this to have been fixed here, but at least please open a ticket for it.

src/backend/mpi/swdsm.cpp
```diff
@@ -419,85 +412,10 @@ void handler(int sig, siginfo_t *si, void *unused){
...
-			loadline[0]=GLOBAL_NULL;
-			loadtag[0]=GLOBAL_NULL;
-#else
-			sem_wait(&loadwaitsem);
-			loadline[0]=GLOBAL_NULL;
-			loadtag[0]=GLOBAL_NULL;
-			prefetchline[0]=GLOBAL_NULL;
-			prefetchtag[0]=GLOBAL_NULL;
-
+		prefetch_cache_entry((alignedDistrAddr+CACHELINE*pagesize), ((startIndex+CACHELINE)%cachesize));
```
SakalisC: Would `(startIndex%cachesize) + 1` be more understandable? This is a minor thing, not really a blocking issue.

src/backend/mpi/swdsm.cpp
```diff
@@ -594,346 +512,319 @@ unsigned long getOffset(unsigned long addr){
...
-		sem_post(&writerwaitsem);
 	}
-	return nullptr;
+	writebufferstart = (writebufferstart+1)%writebuffersize;
+	sem_post(&ibsem);
 }
 
-void * loadcacheline(void * x){
-	UNUSED_PARAM(x);
+void load_cache_entry(unsigned long loadtag, unsigned long loadline) {
```
SakalisC: Same comment about naming here.

src/backend/mpi/swdsm.cpp
```diff
@@ -594,346 +512,319 @@ unsigned long getOffset(unsigned long addr){
...
+	if(tmptag == lineAddr && tmpstate != INVALID){
+		sem_post(&ibsem);
+		return;
+	}
 
 
-		void * lineptr = (char*)startAddr + lineAddr;
+	void * lineptr = (char*)startAddr + lineAddr;
 
+	if(cacheControl[startidx].tag  != lineAddr){
```
SakalisC: What's happening here? Why are there two if statements, one after the other, with the same condition?

src/backend/mpi/swdsm.cpp
```diff
@@ -594,346 +512,319 @@ unsigned long getOffset(unsigned long addr){
...
+	if(tmptag == lineAddr && tmpstate != INVALID){ //trying to load already valid ..
+		sem_post(&ibsem);
+		return;
+	}
 
 
-		void * lineptr = (char*)startAddr + lineAddr;
+	void * lineptr = (char*)startAddr + lineAddr;
 
+	if(cacheControl[startidx].tag  != lineAddr){
```
SakalisC: same duplicated if statement here

SakalisC:
Seems fine to me, I only have minor comments. There are a number of things that need to be fixed in this code, but I guess they can be left for future work. Since most of the code is the same, I skimmed over the code that looked like pre-existing code but with different indentation.

src/backend/mpi/swdsm.cpp
davidklaftenegger: I think this one is important to discuss a little bit: should this even be two parameters?
Note that startIndex (which seeds line) is essentially computed from the alignedDistrAddr (which seeds tag).
See definitions /  computation at
https://github.com/davidklaftenegger/argodsm/blob/ebea0f3cd8e4cf5623b7e362101e36e6aa2df0c0/src/backend/mpi/swdsm.cpp#L317-L324
https://github.com/davidklaftenegger/argodsm/blob/ebea0f3cd8e4cf5623b7e362101e36e6aa2df0c0/src/backend/mpi/swdsm.cpp#L274-L277
and usage at
https://github.com/davidklaftenegger/argodsm/blob/ebea0f3cd8e4cf5623b7e362101e36e6aa2df0c0/src/backend/mpi/swdsm.cpp#L415-L418


Maybe we should restructure this and only have a single parameter, which could either be the aligned or unaligned address of the miss. Opinions?
CC @magnusandersnorgren @pekemark @lundgren87 @kostis

src/backend/mpi/swdsm.cpp
davidklaftenegger: Given that CACHELINE is defined as 1L, your suggestion would allow the entry to go out of bounds, no?

src/backend/mpi/swdsm.cpp
davidklaftenegger: If you want historical explanations please ask @magnusandersnorgren - I'd assume remnants of debugging.
Otherwise I'll interpret this as "please remove superfluous if()".

src/backend/mpi/swdsm.cpp
SakalisC: Right, my bad

src/backend/mpi/swdsm.cpp
SakalisC: Let's go with "please remove superfluous if()"

src/backend/mpi/swdsm.cpp
SakalisC: As we talked in private, maybe such changes should be left for a future PR.

src/backend/mpi/swdsm.cpp
```diff
@@ -326,8 +321,6 @@ void handler(int sig, siginfo_t *si, void *unused){
 
 	unsigned long * localAlignedAddr = (unsigned long *)((char*)startAddr + lineAddr);
 	unsigned long startIndex = getCacheIndex(lineAddr);
-	unsigned long cacheIndex = getCacheIndex(alignedDistrAddr);
-	cacheIndex = startIndex;
 	alignedDistrAddr = lineAddr;
```
pekemark: Not relevant to this PR, but this alias needs to disappear.

src/backend/mpi/swdsm.cpp
pekemark: The difference between `load_cache_entry` and `prefetch_cahce_entry` is actually extremely minimal. The only functional difference I can find between them is that the former performs the test `tempsharer != id` in one of the if-cases, while the later does not perform the test in the equivalent if-case.

davidklaftenegger: related to #28?

src/backend/mpi/swdsm.cpp
SakalisC: I will leave it up to you if you want to do this in this PR or leave it for later. Since it's not a functional change I am going to go ahead and not care.

src/backend/mpi/swdsm.cpp
pekemark: I've been looking a bit at the relevant if-statement. My conclusion is that the `tempsharer != id` condition is not strictly necessary, the side-effects of the if-case will be skipped anyway should the condition be true. Someone should double check this however.

src/backend/mpi/swdsm.cpp
davidklaftenegger: This issue is now tracked at #31

src/backend/mpi/swdsm.cpp
davidklaftenegger: This issue is now tracked in #32

src/backend/mpi/swdsm.cpp
davidklaftenegger: This will be resolved together with #32, as the entire code will probably change.

src/backend/mpi/swdsm.cpp
davidklaftenegger: again, see #32

src/backend/mpi/swdsm.cpp
davidklaftenegger: again, see #31

davidklaftenegger:
rebased to current master, no other changes. also awaiting final approval from jenkins.

src/backend/mpi/swdsm.cpp
davidklaftenegger: This issue is addressed in #34

PR #27, opened 2019-11-21
Ensure that lockWindow and workcomm are freed

lundgren87:
The latest version of Intel MPI allows master to compile without
this fix, but will fail in MPI_Finalize due to "Device or resource
busy". Ensuring that lockWindow and workcomm are properly freed
before calling MPI_Finalize fixes this issue.

### DIFF
--- a/src/backend/mpi/swdsm.cpp
+++ b/src/backend/mpi/swdsm.cpp
@@ -1215,6 +1215,8 @@ void argo_finalize(){
 		MPI_Win_free(&globalDataWindow[i]);
 	}
 	MPI_Win_free(&sharerWindow);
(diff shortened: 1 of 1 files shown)


kostis:
There has not been much reaction on this.  Except from the difference in style (tabs vs spaces) of the patch, that perhaps should be corrected. it looks good to me and can be merged.

Also, perhaps we should make sure that Jason has the latest version of Intel MPI installed so that we test this.

lundgren87:
> There has not been much reaction on this. Except from the difference in style (tabs vs spaces) of the patch, that perhaps should be corrected. it looks good to me and can be merged.
This should now be fixed in line with the ArgoDSM code standard.

PR #34, opened 2020-02-25
handler() variable redundancy/naming cleanup

davidklaftenegger:
This patch cleans up some redundant variables and confusing naming in the segfault handler function.


davidklaftenegger:
hmm. github presents this in a confusing way. This should only be one commit (the last one).

PR #35, opened 2020-02-25
handler() variable redundancy/naming cleanup

davidklaftenegger:
This patch cleans up some redundant variables and confusing naming in the segfault handler function.

Due to my inability to use github nicely, this replaces #34

### DIFF
--- a/src/backend/mpi/swdsm.cpp
+++ b/src/backend/mpi/swdsm.cpp
@@ -3,6 +3,8 @@
  * @brief This file implements the MPI-backend of ArgoDSM
  * @copyright Eta Scale AB. Licensed under the Eta Scale Open Source License. See the LICENSE file for details.
  */
+#include<cstddef>
+
 #include "signal/signal.hpp"
 #include "virtual_memory/virtual_memory.hpp"
 #include "swdsm.h"
@@ -295,14 +297,14 @@ void init_mpi_cacheblock(void){
 	MPI_Type_commit(&cacheblock);
 }
 
-unsigned long alignAddr(unsigned long addr){
-	unsigned long mod = addr % pagesize;
-	if(addr % pagesize != 0){
-		addr = addr - mod;
-	}
-	addr /= pagesize;
-	addr *= pagesize;
-	return addr;
+/**
+ * @brief align an offset into a memory region to the beginning of its size block
+ * @param offset the unaligned offset
+ * @param size the size of each block
+ * @return the beginning of the block of size size where offset is located
+ */
+inline std::size_t align_backwards(std::size_t offset, std::size_t size) {
+	return (offset / size) * size;
 }
 
 void handler(int sig, siginfo_t *si, void *unused){
@@ -312,19 +314,18 @@ void handler(int sig, siginfo_t *si, void *unused){
 
 	unsigned long tag;
 	argo_byte owner,state;
-	unsigned long distrAddr =  (unsigned long)((unsigned long)(si->si_addr) - (unsigned long)(startAddr));
+	/* compute offset in distributed memory in bytes, always positive */
+	const std::size_t access_offset = static_cast<char*>(si->si_addr) - static_cast<char*>(startAddr);
 
-	unsigned long alignedDistrAddr = alignAddr(distrAddr);
-	unsigned long remCACHELINE = alignedDistrAddr % (CACHELINE*pagesize);
-	unsigned long lineAddr = alignedDistrAddr - remCACHELINE;
-	unsigned long classidx = get_classification_index(lineAddr);
+	/* align access offset to cacheline */
+	const std::size_t aligned_access_offset = align_backwards(access_offset, CACHELINE*pagesize);
+	unsigned long classidx = get_classification_index(aligned_access_offset);
 
-	unsigned long * localAlignedAddr = (unsigned long *)((char*)startAddr + lineAddr);
-	unsigned long startIndex = getCacheIndex(lineAddr);
-	alignedDistrAddr = lineAddr;
-
-	unsigned long homenode = getHomenode(lineAddr);
-	unsigned long offset = getOffset(lineAddr);
+	/* compute start pointer of cacheline. char* has byte-wise arithmetics */
+	char* const aligned_access_ptr = static_cast<char*>(startAddr) + aligned_access_offset;
+	unsigned long startIndex = getCacheIndex(aligned_access_offset);
+	unsigned long homenode = getHomenode(aligned_access_offset);
+	unsigned long offset = getOffset(aligned_access_offset);
 	unsigned long id = 1 << getID();
 	unsigned long invid = ~id;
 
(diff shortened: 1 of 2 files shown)


src/backend/mpi/swdsm.cpp
```diff
@@ -312,19 +314,18 @@ void handler(int sig, siginfo_t *si, void *unused){
...
+	std::size_t aligned_access_offset = align_backwards(access_offset, CACHELINE*pagesize);
+	unsigned long classidx = get_classification_index(aligned_access_offset);
+
+	/* compute start pointer of cacheline. char* has byte-wise arithmetics */
+	char* aligned_access_ptr = reinterpret_cast<char*>(startAddr) + aligned_access_offset;
+	unsigned long startIndex = getCacheIndex(aligned_access_offset);
+	unsigned long homenode = getHomenode(aligned_access_offset);
+	unsigned long offset = getOffset(aligned_access_offset);
 	unsigned long id = 1 << getID();
 	unsigned long invid = ~id;
```
lundgren87: Eventually we should ensure that _all_ variables have better descriptive and more consistent names, in addition to using snake_case in accordance with the code standard. Not important for this specific PR, but the simple fact that `id` equals `1 << getID()` rather than simply the return value of `getID()` is not straightforward.

src/backend/mpi/swdsm.cpp
```diff
@@ -295,14 +297,14 @@ void init_mpi_cacheblock(void){
...
-	addr /= pagesize;
-	addr *= pagesize;
-	return addr;
+/**
+ * @brief align an offset into a memory region to the beginning of its size block
+ * @param offset the pointer or offset
+ * @param size the size of each block
+ * @return the beginning of the block of size size where offset is located
+ */
+inline std::size_t align_backwards(std::size_t offset, std::size_t size) {
```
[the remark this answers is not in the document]
davidklaftenegger: In this particular instance I believe it is the correct type. I have considered uint64_t and std::ptrdiff_t, but the former is quite nondescript and the latter is a signed type. In this particular instance we are dealing with an offset into a memory region, which can only be positive (and the later code complains a lot if the value is signed).

src/backend/mpi/swdsm.cpp
```diff
@@ -312,19 +314,18 @@ void handler(int sig, siginfo_t *si, void *unused){
...
-	unsigned long classidx = get_classification_index(lineAddr);
-
-	unsigned long * localAlignedAddr = (unsigned long *)((char*)startAddr + lineAddr);
-	unsigned long startIndex = getCacheIndex(lineAddr);
-	alignedDistrAddr = lineAddr;
-
-	unsigned long homenode = getHomenode(lineAddr);
-	unsigned long offset = getOffset(lineAddr);
+	/* compute offset in distributed memory in bytes, always positive */
+	std::size_t access_offset = static_cast<char*>(si->si_addr) - static_cast<char*>(startAddr);
```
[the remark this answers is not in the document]
davidklaftenegger: I don't think we have a policy on it; what would be the advantage of having these `const`? Should we have a discussion someday about making a lot more things `const`?

src/backend/mpi/swdsm.cpp
davidklaftenegger: Sure, eventually I would like to get rid of everything in this file and replace it with more modular, easier to read, better documented, and more consistent code. Should we create another ticket for this particular complaint, in the hope someone addresses it?

src/backend/mpi/swdsm.cpp
```diff
@@ -312,19 +314,18 @@ void handler(int sig, siginfo_t *si, void *unused){
...
+
+	/* align access offset to cacheline */
+	std::size_t aligned_access_offset = align_backwards(access_offset, CACHELINE*pagesize);
+	unsigned long classidx = get_classification_index(aligned_access_offset);
+
+	/* compute start pointer of cacheline. char* has byte-wise arithmetics */
+	char* aligned_access_ptr = reinterpret_cast<char*>(startAddr) + aligned_access_offset;
```
davidklaftenegger: Why is none of you complaining about this `reinterpret_cast` that should be a `static_cast`?

src/backend/mpi/swdsm.cpp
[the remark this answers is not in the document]
SakalisC: Mostly as a way of catching unexpected bugs, if we do not except these to be changed anywhere in the code.

src/backend/mpi/swdsm.cpp
SakalisC: I didn't notice :)

src/backend/mpi/swdsm.cpp
lundgren87: I believe it is out of the scope of this PR, as changing either of the variable names above would lead to changes on multiple locations in the backend. Sounds like a good idea for another ticket to be dealt with in the future.

src/backend/mpi/swdsm.cpp
```diff
@@ -411,10 +412,10 @@ void handler(int sig, siginfo_t *si, void *unused){
 	state  = cacheControl[startIndex].state;
 	tag = cacheControl[startIndex].tag;
 
-	if(state == INVALID || (tag != lineAddr && tag != GLOBAL_NULL)){
-		load_cache_entry(alignedDistrAddr, (startIndex%cachesize));
+	if(state == INVALID || (tag != aligned_access_offset && tag != GLOBAL_NULL)) {
+		load_cache_entry(aligned_access_offset, (startIndex%cachesize));
 #ifdef DUAL_LOAD
-		prefetch_cache_entry((alignedDistrAddr+CACHELINE*pagesize), ((startIndex+CACHELINE)%cachesize));
+		prefetch_cache_entry((aligned_access_offset+CACHELINE*pagesize), ((startIndex+CACHELINE)%cachesize));
```
SakalisC: I notice that `alignedDistrAddr` and `lineAddr` seem to have been merged in the new code, I assume it's on purpose?

davidklaftenegger: see line 324 (before the patch): `alignedDistrAddr = lineAddr;`
This is the alias that @pekemark complained about in #26

src/backend/mpi/swdsm.cpp
davidklaftenegger: This is now issue #36, please feel free to contribute your find there.

src/backend/mpi/swdsm.cpp
davidklaftenegger: fixed

src/backend/mpi/swdsm.cpp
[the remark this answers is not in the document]
davidklaftenegger: accepted

src/backend/mpi/swdsm.cpp
SakalisC: Ah right, I missed that.

PR #37, opened 2020-02-27
build shared libraries

### LINKED ISSUE #16: Insufficient linking with librt
### Issue
Compiling and linking external projects with ArgoDSM produces the following error.
`undefined reference to symbol 'shm_unlink@@GLIBC_2.2.5'`
`//lib/x86_64-linux-gnu/librt.so.1: error adding symbols: DSO missing from command line`

shm_open/shm_unlink [requires linking with librt](http://man7.org/linux/man-pages/man3/shm_open.3.html) (-lrt). This is currently done on src/CMakeLists.txt line 37 through `set(vm_libs rt)` but this does not appear to be sufficient when the ArgoDSM libraries are linked from an external project.

### Steps to reproduce

1. Follow the ArgoDSM [quickstart guide](https://etascale.github.io/argodsm/) to build ArgoDSM.
2. Run make install.
3. Follow the ArgoDSM [tutorial](https://etascale.github.io/argodsm/tutorial.html) to compile and link argo_example.cpp with the ArgoDSM libraries, reproducing the error.
4. Link with librt by adding -lrt to the final line in order to successfully compile. (`-largo -largobackend-mpi -lrt`)

Tested with gcc/g++ (mpicc/mpic++) versions 6.5.0 and 7.3.0.

davidklaftenegger:
When using VM_SHM, ArgoDSM requires to link against rt, which we already specify in `CMakeLists.txt`.
However, for static libraries this is not recorded in the final library file, while for dynamically linked libraries it is. This means using ArgoDSM becomes easier if we build only shared libraries. It also avoids similar issues in the future.

To track that compiling works as specified in our tutorial, a corresponding check has been added to jenkins.

Please only approve this PR after testing it a bit, we do not want to break user code with this change.

Fixes #16

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -130,7 +130,7 @@ make install
 
 This step will copy the final ArgoDSM include files to `/usr/local/include` and
 libraries to `/usr/local/lib/`. You can choose a different path above if you
-want, but remember to either set `INCLUDE_PATH`, `LIBRARY_PATH`, and
+want, but remember to either set `LIBRARY_PATH`, `INCLUDE_PATH`, and
 `LD_LIBRARY_PATH` accordingly, or provide the correct paths (`-L`, `-I`, and
 `-Wl,-rpath,`) to your compiler when compiling applications for ArgoDSM.
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -30,6 +30,8 @@ respectively are required.
 
 ### Building ArgoDSM
 
+Note: adjust the below commands to your needs, especially `CMAKE_INSTALL_PREFIX`.
+
 ``` bash
 git clone https://github.com/etascale/argodsm.git
 cd argodsm
@@ -40,15 +42,17 @@ mv googletest-release-1.7.0 gtest-1.7.0
 cd ..
 mkdir build
 cd build
-cmake -DARGO_BACKEND_MPI=ON        \
-      -DARGO_BACKEND_SINGLENODE=ON \
-      -DARGO_TESTS=ON              \
-      -DBUILD_DOCUMENTATION=ON     \
-      -DCMAKE_CXX_COMPILER=mpic++  \
-      -DCMAKE_C_COMPILER=mpicc     \
+cmake -DARGO_BACKEND_MPI=ON              \
+      -DARGO_BACKEND_SINGLENODE=ON       \
+      -DARGO_TESTS=ON                    \
+      -DBUILD_DOCUMENTATION=ON           \
+      -DCMAKE_CXX_COMPILER=mpic++        \
+      -DCMAKE_C_COMPILER=mpicc           \
+      -DCMAKE_INSTALL_PREFIX=/usr/local/ \
       ../
 make
 make test
+make install
 ```
 
 Initially, you need to get the ArgoDSM sources. If you already have a tarball,
@@ -86,16 +90,18 @@ the `ccmake` tool. The difference is that the first one accepts all the build
 options as command line arguments, while the second one works interactively.
 Below is an example call to `cmake` with all the recommended command line
 arguments. If you plan on contributing to the ArgoDSM source code, you should
-also enable the `ARGO_DEBUG` option. After generating the makefiles, then just
-build the library and executables with a simple make command.
+also enable the `ARGO_DEBUG` option. Remember to change `CMAKE_INSTALL_PREFIX`
+to a path that you have write access to. After generating the makefiles proceed
+to building the library and executables with a simple make command.
 
 ``` bash
-cmake -DARGO_BACKEND_MPI=ON        \
-      -DARGO_BACKEND_SINGLENODE=ON \
-      -DARGO_TESTS=ON              \
-      -DBUILD_DOCUMENTATION=ON     \
-      -DCMAKE_CXX_COMPILER=mpic++  \
-      -DCMAKE_C_COMPILER=mpicc     \
+cmake -DARGO_BACKEND_MPI=ON              \
+      -DARGO_BACKEND_SINGLENODE=ON       \
+      -DARGO_TESTS=ON                    \
+      -DBUILD_DOCUMENTATION=ON           \
+      -DCMAKE_CXX_COMPILER=mpic++        \
+      -DCMAKE_C_COMPILER=mpicc           \
+      -DCMAKE_INSTALL_PREFIX=/usr/local/ \
       ../
 make
 ```
@@ -117,6 +123,16 @@ Keep in mind that currently even the MPI tests are run on a single node without
 invoking `mpirun`. After running the previous command, you should also run the
 tests with at least 4 nodes. To learn how to run the MPI tests on multiple
 nodes, proceed to the next section.
+
+``` bash
+make install
+```
+
+This step will copy the final ArgoDSM include files to `/usr/local/include` and
+libraries to `/usr/local/lib/`. You can choose a different path above if you
+want, but remember to either set `INCLUDE_PATH`, `LIBRARY_PATH`, and
+`LD_LIBRARY_PATH` accordingly, or provide the correct paths (`-L`, `-I`, and
+`-Wl,-rpath,`) to your compiler when compiling applications for ArgoDSM.
 
 ### Running the Applications
 
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -176,17 +176,18 @@ The final result of all the changes is this:
 {% include argo_example.cpp %}
 ```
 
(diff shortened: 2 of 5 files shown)


davidklaftenegger:
> It is worth noting that it is now no longer sufficient to _just_ include the library path when compiling external applications (`-L${ARGO_INSTALL_DIRECTORY}/lib`), but that the ArgoDSM library files now _must_ be present in your `$LD_LIBRARY_PATH` in order to be found at runtime.

This is not the case in my test, and it should not be the case. Have you used absolute paths in both cases? I am using the following command to compile successfully:
```
mpic++ -o test_compiling_example -O3 -std=c++11 -L$(pwd)/../install/lib -I$(pwd)/../install/include ../argodsm/docs/_includes/argo_example.cpp -largo -largobackend-mpi
```

lundgren87:
> > It is worth noting that it is now no longer sufficient to _just_ include the library path when compiling external applications (`-L${ARGO_INSTALL_DIRECTORY}/lib`), but that the ArgoDSM library files now _must_ be present in your `$LD_LIBRARY_PATH` in order to be found at runtime.
> 
> This is not the case in my test, and it should not be the case. Have you used absolute paths in both cases? I am using the following command to compile successfully:
> 
> ```
> mpic++ -o test_compiling_example -O3 -std=c++11 -L$(pwd)/../install/lib -I$(pwd)/../install/include ../argodsm/docs/_includes/argo_example.cpp -largo -largobackend-mpi
> ```

**mpic++** output:
```
mpic++ -o paratest -O3 -std=c++11 -L/home/sven/git/argodsm/install/lib -I/home/sven/git/argodsm/install/include  paratest.cpp -largo -largobackend-mpi
```
**ldd** output:
```
	linux-vdso.so.1 (0x00007ffdb6daa000)
	libargo.so => not found
	libargobackend-mpi.so => not found
	libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f59e6ab2000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f59e6a98000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f59e6a75000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f59e6884000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f59e6733000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f59e6cc1000)
```
**echo $LD_LIBRARY_PATH** output:
```
:/usr/local/lib
```
ArgoDSM is installed through adding `DCMAKE_INSTALL_DIR=/home/sven/git/argodsm/install` and running `make install`, and all files are located where they should be (in the install directory).
Unless I add the library directory to my `LD_LIBRARY_PATH`, I get the following error when running with mpirun:
`./paratest: error while loading shared libraries: libargo.so: cannot open shared object file: No such file or directory`
Same issue on Jason, with slightly different ldd output. What am I missing?

davidklaftenegger:
>
>
> > > It is worth noting that it is now no longer sufficient to _just_ include the library path when compiling external applications (`-L${ARGO_INSTALL_DIRECTORY}/lib`), but that the ArgoDSM library files now _must_ be present in your `$LD_LIBRARY_PATH` in order to be found at runtime.
> >
> > This is not the case in my test, and it should not be the case. Have you used absolute paths in both cases? I am using the following command to compile successfully:
> ```
> 	libargo.so => not found
> 	libargobackend-mpi.so => not found
> ```

> Same issue on Jason, with slightly different ldd output. The libraries and headers are clearly found while compiling but not at runtime, hence the need for setting `LD_LIBRARY_PATH`. As far as I am aware this is standard procedure, what am I missing?
> 
> **EDIT**: Another way to solve it is to provide mpic++ with an rpath (`-Wl,-rpath,'path/to/argo/lib`), but the only rpath mpic++ provides by default is to its own dependencies.

I thought the initial comment referred to compilation, not runtime (which of course makes no sense considering LD_LIBRARY_PATH _only_ affects runtime).

Notably, when building with CMake it appears to add the correct linker `rpath` flags for me, but you are correct that this or setting LD_LIBRARY_PATH is required.

You are also correct that the tutorial should make this more obvious. I'll look at it and amend this pull-request accordingly.

docs/index.md
```diff
@@ -40,15 +40,17 @@ mv googletest-release-1.7.0 gtest-1.7.0
...
+      -DARGO_BACKEND_SINGLENODE=ON       \
+      -DARGO_TESTS=ON                    \
+      -DBUILD_DOCUMENTATION=ON           \
+      -DCMAKE_CXX_COMPILER=mpic++        \
+      -DCMAKE_C_COMPILER=mpicc           \
+      -DCMAKE_INSTALL_PREFIX=/usr/local/ \
       ../
 make
 make test
+make install
```
SakalisC: I am not sure I like the idea of having the `make install` in the basic instructions since, presumably, more users will not have root permissions on their cluster. I might be wrong about the "most users" part though. What is the common practice for open source applications?

davidklaftenegger: Given current usage, I think "most users" will replace the path with one in their home directory. @lundgren87 can probably explain better why they started using `make install` to begin with, but it appears that using stuff from the build tree is less standard practice than changing install paths and `make install`.

From my limited experience the default prefix seems to be `/usr/local/` in most projects, even when root privileges are not required for the software. The documentation patch already adds the option that needs to be changed to change the path (4 lines above the quoted line), do you think it should be phrased in a way that nudges people into actually changing it?

SakalisC: Maybe adding something along the lines of "Note that you might need root access to install to the /usr/local/ directory. If you do not have permission to do this on your system, you should change the parameter to a directory where you have access to. Please contact your system administrator for more information."?

Maybe without the "contact your admin line", since it's already mentioned somewhere else.

kostis: Including a `make install` is sort of standard practice in most code bases and the default prefix is indeed `/usr/local`.  Most code bases are mute about permissions or `sudo`.  The only difference is that most code bases I am aware of also include a `configure` step where one can specify an appropriate `--prefix` (if `/usr/local` is not the desired one).  See for example [CVC4's instructions](https://github.com/CVC4/CVC4/blob/master/INSTALL.md).

Rather than including the verbose text of @SakalisC, I suggest we just make sure that the instructions explicitly mention how one specifies a different prefix.

davidklaftenegger: see new version 4a26f57

docs/index.md
```diff
@@ -30,6 +30,8 @@ respectively are required.
 
 ### Building ArgoDSM
 
+Note: adjust thte below commands to your needs, especially `CMAKE_INSTALL_PREFIX`.
```
SakalisC: There is a typo here.

docs/index.md
```diff
@@ -118,6 +124,16 @@ invoking `mpirun`. After running the previous command, you should also run the
...
 nodes, proceed to the next section.
 
+``` bash
+make install
+```
+
+This step will copy the final ArgoDSM include files to `/usr/local/include` and
+libraries to `/usr/local/lib/`. You can choose a different path above if you
+want, but remember to either set `INCLUDE_PATH`, `LIBRARY_PATH`, and
+`LD_LIBRARY_PATH` accordingly, or provide the correct paths (`-L`, `-I`, and
```
pekemark: The `-L` and `-I` flags should swap places to match the list of environment variables. (Or the environment variables should change order.)

docs/tutorial.md
```diff
@@ -176,17 +176,18 @@ The final result of all the changes is this:
...
 
-We can now compile the application. We assume that `${ARGO_INSTALL_DIRECTORY}` is
-where you installed ArgoDSM. If the ArgoDSM files are already in your
-`(LD_)LIBRARY_PATH` and include path, you can skip the `-L...` and `-I...`
-switches. If you have no idea what we are talking about, you should ask your
-system administrator.
+We can now compile the application. We assume that `${ARGO_INSTALL_DIRECTORY}`
+is where you installed ArgoDSM. If the ArgoDSM files are already in your
+`LIBRARY_PATH`, `INCLUDE_PATH`, and `LD_LIBRARY_PATH`, you can skip the
+`-L...`, `-I...`, and `-Wl,-rpath,...` switches. If you have no idea what we
```
pekemark: The order of the environment variables and compiler flags should ideally match the comment in docs/index.md. I'm not particular about which place changes.

pekemark: Order altered to match in docs/index.md by @davidklaftenegger.

kostis:
I will go ahead and merge this.

PR #38, opened 2020-03-26
Run tests for the MPI backend through mpirun

lundgren87:
This pull request ensures that all tests for the MPI backend are run through mpirun. By default, each test is run using _two_ ArgoDSM software nodes. This can be changed by setting the `ARGO_TESTS_NPROCS` CMake option to any number between 1 and 8.

The purpose of this is change is to expose the whole backend during tests, as running on at least two nodes is required in order to ensure that the `handler()`, `load_cache_entry()` and `prefetch_cache_entry()` are fully executed at some point.

In order to still be able to run `make test` on most machines using multiple software nodes, the ArgoDSM cache size for the tests using 256MB or more memory is significantly reduced in order to avoid allocating too much local memory for the cache. Additionally, in order to avoid cluttering the output file (`Testing/Temporary/LastTest.log`), output is suppressed for all nodes besides node 0.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -119,10 +119,15 @@ library, as well as *exactly* one of the backend libraries.
 make test
 ```
 
-Keep in mind that currently even the MPI tests are run on a single node without
-invoking `mpirun`. After running the previous command, you should also run the
-tests with at least 4 nodes. To learn how to run the MPI tests on multiple
-nodes, proceed to the next section.
+This step executes the ArgoDSM tests. Tests for the MPI backend are run on two
+ArgoDSM software nodes by default. This number can be changed through setting
+the `ARGO_TESTS_NPROCS` CMake option to any number from 1 to 8. Keep in mind
+that some MPI distributions may not allow executing more processes than the
+number of physical cores in the system. Please note that some issues are only
+encountered when running with more than two ArgoDSM nodes or on multiple
+hardware nodes. For this reason it is highly recommended to run the MPI tests
+on at least four hardware nodes when possible. Refer to the next section to
+learn how to run applications on multiple hardware nodes.
 
 ``` bash
 make install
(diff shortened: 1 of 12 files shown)


docs/index.md
```diff
@@ -119,10 +119,11 @@ library, as well as *exactly* one of the backend libraries.
...
 ```
 
-Keep in mind that currently even the MPI tests are run on a single node without
-invoking `mpirun`. After running the previous command, you should also run the
-tests with at least 4 nodes. To learn how to run the MPI tests on multiple
-nodes, proceed to the next section.
+This step executes the ArgoDSM tests. Tests for the MPI backend are run on two
+ArgoDSM software nodes by default. This number can be changed through setting
+the `ARGO_TESTS_NPROCS` CMake option to any number from 1 to 8. Keep in mind
+that some MPI distributions may not allow executing more processes than your
```
kostis, line 125: I suggest: ... may not allow executing more processes than the number of physical cores in the system.

lundgren87:
> Looks good except from the fact that it's not clear what the 8 and 2 numbers that divide the `size` in the second argument of `argo::init` denote (and why they appear there). Can't we have some appropriate name for this divisor?
> 
> Also, for uniformity, should we change all calls to `argo::init` to be with two arguments?

https://github.com/etascale/argodsm/blob/de343a833645552023195b797300a7a73870e6f7/src/argo.hpp#L54
I can amend the PR and call this parameter `cache_size` to make this more clear. The reason I believe the cache size should be significantly smaller for most tests is in order to allow running `make test` on machines with limited memory using multiple software nodes. Several tests currently use 256MB ArgoDSM memory, and executing one test on eight software nodes with 256MB memory and 256MB cache size will require over 2GB memory (9*256MB).

I believe the way to call `argo::init` is appropriately done using both parameters, but there are default values in place of `argo_size` 8GB and `cache_size` 1GB should no value be specified through either the init call or the use of environment variables.

docs/index.md
lundgren87: This should now be reflected in the latest update.

kostis:
Looks good to me now.  
@davidklaftenegger will also look at this soon and then we can squash + merge.

lundgren87:
The latest update now no longer disables test output for processes above rank 0 in order to ensure that no unexpected bugs slip through. Additionally, the quickstart guide now again mentions the importance of testing on multiple hardware nodes.

kostis:
Unless there is some objection, I will squash and merge this tonight.

Issue #32, opened 2020-02-24
prefetch_cache_entry is redundant


davidklaftenegger:
The `prefetch_cache_entry` function is largely redundant.

Considering that @lundgren87 is working on a better algorithm for this functionality no action is needed until the algorithm has been replaced, but this issue is used to track the issue of `prefetch_cache_entry`in case that update gets delayed.

Issue #30, opened 2020-01-17
Remote load does not add self as sharer in local Pyxis directory


lundgren87:
When loading or prefetching an ArgoDSM page from another ArgoDSM node for the first time, the remote Pyxis directory is updated with the fetching node as sharer **but the local Pyxis directory is not.**
https://github.com/etascale/argodsm/blob/ef644145ad85955cb229b962b0f5e6445349bb94/src/backend/mpi/swdsm.cpp#L738-L741
`tempsharer` _contains the remote Pyxis directory content **before** adding the fetching node as sharer._

This causes the update of the local Pyxis directory to be deferred until the next read miss on the same page (a write miss will only mark the node as writer, not sharer), which will then recognize itself as sharer due to reading the remote Pyxis directory.

Any synchronization points (performing self-invalidation) between the first and second read miss on a page (classification index to be exact) will cause the page to erroneously fail the _No writer and assert that the node is a sharer_ check as it is not locally recognized as sharer yet.
https://github.com/etascale/argodsm/blob/ef644145ad85955cb229b962b0f5e6445349bb94/src/backend/mpi/swdsm.cpp#L1243-L1255
The page will therefore go on to be invalidated from the cache.

Fixing this is simple (just add `globalSharers[classidx] |= id;` at appropriate location) but needs testing for any unexpected side effects.

Issue #29, opened 2020-01-17
Pyxis directory size scaling with cache_size instead of argo_size


lundgren87:
In the current implementation, the **Pyxis directory size** is set to:
https://github.com/etascale/argodsm/blob/ef644145ad85955cb229b962b0f5e6445349bb94/src/backend/mpi/swdsm.cpp#L1045
and the classification index corresponding to an address is set to:
https://github.com/etascale/argodsm/blob/ef644145ad85955cb229b962b0f5e6445349bb94/src/backend/mpi/swdsm.cpp#L1425-L1427
When using a `cache_size` significantly smaller than `argo_size`, this means that multiple pages will share the same classification index and (possibly) homenode will all be considered downgraded if one page is, resulting in possible performance degradation from unnecessary invalidations.

Is this intended behavior in order to minimize the Pyxis directory size, or is this simply a relic from the time that `cache_size` was not set by user input?

The amount (or size) of remote operations performed by ArgoDSM does not scale with directory size as far as I can tell, and allowing the directory size to scale with argo_size (at least to some limit) _should be positive in terms of performance_. However, for **very** large memory allocations the directory size can possibly be an issue, as for most linux platforms the directory size will be `argo_size/256`.

Issue #28, opened 2019-12-12
Rare coherence failure when prefetching a page unused until after synchronization


lundgren87:
### Issue
A rare coherence failure can occur when a remote ArgoDSM page is prefetched to a node at one point, but not accessed until after another node has released changes made to the page. ArgoDSM may during synchronization falsely consider the prefetched page to be in "single writer" state, and keep it on the node (without performing a self-downgrade on it) instead of correctly self-invalidating the page. This error does not correct itself unless the page is evicted from the node cache, and subsequent reads from this page will not recognize changes made by remote nodes even after global synchronization.
**Note that this issue only occurs if prefetch (DUAL_LOAD) is enabled.**

### How to reproduce
It is possible to reliably reproduce this error by initializing ArgoDSM with specific argo_size and cache_size parameters along with specific code to trigger the offending prefetch (to ensure that a page is prefetched but not used). I have been able to reproduce this with a cache_size smaller than argo_size/2 in the following code:
https://github.com/lundgren87/paratest/blob/prefetch_fail/simpletest.cpp

This specific code must executed on **exactly two ArgoDSM nodes** (same result obtained on HW or SW nodes) in order to reproduce this failure (other values of argo_size and cache_size may reproduce the issue with other node sizes) as such:
`mpirun -n 2 ./simpletest`

### ArgoDSM details
- In the example above, 20 ArgoDSM pages are allocated during initialization, pages 0-9 on "node 0" and 10-19 on "node 1". Node 0 initializes pages 9-12, which triggers remote loads of pages 10 and 12, and prefetch of pages 11 and 13.  Page 13 is not accessed during initialization by node 0. Node 1 meanwhile initializes pages 13-16.
- At the first synchronization point, page 13 is considered to be in "single writer" state by node 0 even though it has not accessed the page and is as such kept on the node (without being self-downgraded, since it is in state CLEAN).
- After the first synchronization point, node 1 writes additional changes to page 13.
- At the second synchronization point, the changes made by node 1 should be made visible to node 0. However, node 0 does not recognize that any writer has been added to page 13 and again keeps it in the cache instead of self-invalidating it.
- After the second synchronization point, node 0 attempts to check writes made to pages 9-16, but upon accessing page 13 it accesses the page present in the local node cache instead of remotely loading page 13 containing the latest updates from node 1, resulting in missed writes.

davidklaftenegger:
[...]

PR #39, opened 2020-06-11
Improved loading and prefetching functionality

lundgren87:
This pull request provides an updated and improved remote page loading functionality (`load_cache_entry`) that reduces the amount of remote operations per page loaded. It also enables prefetching of subsequent ArgoDSM pages located on the same remote node for a fraction of the cost compared to loading each page individually. This is achieved by performing both remote Pyxis operations and remote data loading **in bulk** when possible rather than individually for each page.

The total number of pages loaded on each remote load operation defaults to 8 (target page and up to 7 subsequent pages), and can be adjusted by setting the `ARGO_LOAD_SIZE` environment variable to the desired number.

As a side effect this PR deals with issues #28 (through commit f6b5306 specifically), #29, #30, and #32.

### DIFF
--- a/src/backend/backend.hpp
+++ b/src/backend/backend.hpp
@@ -80,6 +80,15 @@ namespace argo {
 		 */
 		std::size_t global_size();
 
+		/**
+		 * @brief Check if a an address is cached or node local
+		 * @param addr the address to verify cache status on
+		 * @return True if the page is either cached on the node
+		 * or locally backed on the node, otherwise false.
+		 * @warning THIS IS FOR TESTING, DO NOT USE
+		 * @todo THIS SHOULD BE BAKED IN TO A CACHE CLASS
+		 */
+		bool is_cached(void* addr);
 
 		/**
 		 * @brief global memory space address
--- a/src/backend/backend.hpp
+++ b/src/backend/backend.hpp
@@ -213,7 +213,8 @@ namespace argo {
 			/**
 			 * @brief Backend internal type erased atomic store function
 			 * @param desired Pointer to the object that holds the desired value
-			 * @param size == sizeof(*desired)
+			 * @param count Number of element from desired to store
+			 * @param size size of each element in desired
 			 * @param rank Rank of target window to which operation will be performed
 			 * @param disp Displacement from start of window to beginning of target buffer
 			 * @sa store to public window section
@@ -222,11 +223,13 @@ namespace argo {
 			 * @warning For internal use only - DO NOT USE UNLESS YOU KNOW WHAT YOU ARE DOING
 			 */
 			void _store_public_owners_dir(const void* desired,
-				const std::size_t size, const std::size_t rank, const std::size_t disp);
+				const std::size_t size, const std::size_t count,
+				const std::size_t rank, const std::size_t disp);
 
 			/**
 			 * @brief Backend internal type erased atomic store function
 			 * @param desired The desired value to be stored
+			 * @param count Number of element from desired to store
 			 * @param rank Rank of target window to which operation will be performed
 			 * @param disp Displacement from start of window to beginning of target buffer
 			 * @sa store to private window section
@@ -235,7 +238,7 @@ namespace argo {
 			 * @warning For internal use only - DO NOT USE UNLESS YOU KNOW WHAT YOU ARE DOING
 			 */
 			void _store_local_owners_dir(const std::size_t* desired,
-				const std::size_t rank, const std::size_t disp);
+				const std::size_t count, const std::size_t rank, const std::size_t disp);
 
 			/**
 			 * @brief Backend internal type erased atomic store function
@@ -264,7 +267,8 @@ namespace argo {
 			/**
 			 * @brief Backend internal type erased atomic load function
 			 * @param output_buffer Pointer to the memory location where the value of the object should be stored
-			 * @param size == sizeof(*output_buffer)
+			 * @param size Size of an element to load
+			 * @param count Number of element to load
 			 * @param rank Rank of target window to which operation will be performed
 			 * @param disp Displacement from start of window to beginning of target buffer
 			 * @sa load from public window section
@@ -273,11 +277,13 @@ namespace argo {
 			 * @warning For internal use only - DO NOT USE UNLESS YOU KNOW WHAT YOU ARE DOING
 			 */
 			void _load_public_owners_dir(void* output_buffer,
-				const std::size_t size, const std::size_t rank, const std::size_t disp);
+				const std::size_t size, const std::size_t count,
+				const std::size_t rank, const std::size_t disp);
 
 			/**
 			 * @brief Backend internal type erased atomic load function
 			 * @param output_buffer Pointer to the memory location where the value of the object should be stored
+			 * @param count Number of element to load
 			 * @param rank Rank of target window to which operation will be performed
 			 * @param disp Displacement from start of window to beginning of target buffer
 			 * @sa load from private window section
@@ -286,7 +292,7 @@ namespace argo {
 			 * @warning For internal use only - DO NOT USE UNLESS YOU KNOW WHAT YOU ARE DOING
 			 */
 			void _load_local_owners_dir(void* output_buffer,
-				const std::size_t rank, const std::size_t disp);
+				const std::size_t count, const std::size_t rank, const std::size_t disp);
 				
 			/**
 			 * @brief Backend internal type erased atomic load function
--- a/src/backend/backend.hpp
+++ b/src/backend/backend.hpp
@@ -222,8 +222,8 @@ namespace argo {
 			/**
 			 * @brief Backend internal type erased atomic store function
 			 * @param desired Pointer to the object that holds the desired value
-			 * @param count Number of element from desired to store
-			 * @param size size of each element in desired
+			 * @param count Number of elements from desired to store
+			 * @param size Size of each element in desired
 			 * @param rank Rank of target window to which operation will be performed
 			 * @param disp Displacement from start of window to beginning of target buffer
 			 * @sa store to public window section
@@ -238,7 +238,7 @@ namespace argo {
 			/**
 			 * @brief Backend internal type erased atomic store function
 			 * @param desired The desired value to be stored
-			 * @param count Number of element from desired to store
+			 * @param count Number of elements from desired to store
 			 * @param rank Rank of target window to which operation will be performed
 			 * @param disp Displacement from start of window to beginning of target buffer
 			 * @sa store to private window section
@@ -277,7 +277,7 @@ namespace argo {
 			 * @brief Backend internal type erased atomic load function
 			 * @param output_buffer Pointer to the memory location where the value of the object should be stored
 			 * @param size Size of an element to load
-			 * @param count Number of element to load
+			 * @param count Number of elements to load
 			 * @param rank Rank of target window to which operation will be performed
 			 * @param disp Displacement from start of window to beginning of target buffer
 			 * @sa load from public window section
@@ -292,7 +292,7 @@ namespace argo {
 			/**
 			 * @brief Backend internal type erased atomic load function
 			 * @param output_buffer Pointer to the memory location where the value of the object should be stored
-			 * @param count Number of element to load
+			 * @param count Number of elements to load
 			 * @param rank Rank of target window to which operation will be performed
 			 * @param disp Displacement from start of window to beginning of target buffer
 			 * @sa load from private window section
--- a/src/backend/mpi/swdsm.cpp
+++ b/src/backend/mpi/swdsm.cpp
@@ -4,6 +4,7 @@
  * @copyright Eta Scale AB. Licensed under the Eta Scale Open Source License. See the LICENSE file for details.
  */
 #include<cstddef>
+#include<vector>
 
 #include "env/env.hpp"
 #include "signal/signal.hpp"
@@ -17,13 +18,6 @@ namespace vm = argo::virtual_memory;
 namespace sig = argo::signal;
 namespace env = argo::env;
 
-/*Threads*/
-/** @brief Thread loads data into cache */
-pthread_t loadthread1;
-/** @brief Thread loads data into cache with an overlapping request (some parts are done in parallel) */
-pthread_t loadthread2;
-/** @brief Thread writes data remotely if parts of writebuffer */
-pthread_t writethread;
 /** @brief For matching threads to more sensible thread IDs */
 pthread_t tid[NUM_THREADS] = {0};
 
@@ -92,18 +86,10 @@ sem_t ibsem;
 /*Loading and Prefetching*/
 /**
  * @brief load into cache helper function
- * @param tag aligned address to load into the cache
- * @param line cache entry index to use
+ * @param aligned_access_offset memory offset to load into the cache
+ * @pre aligned_access_offset must be aligned as CACHELINE*pagesize
  */
-void load_cache_entry(unsigned long tag, unsigned long line);
-
-/**
- * @brief prefetch into cache helper function, which duplicates a lot of code
- * @param tag aligned address to prefetch into the cache
- * @param line cache entry index to use
- * @todo this function duplicates a lot of code from load_cache_entry(), this should be fixed
- */
-void prefetch_cache_entry(unsigned long tag, unsigned long line);
+void load_cache_entry(std::size_t aligned_access_offset);
 
 /*Global lock*/
 /** @brief  Local flags we spin on for the global lock*/
@@ -354,10 +340,7 @@ void handler(int sig, siginfo_t *si, void *unused){
 	tag = cacheControl[startIndex].tag;
 
 	if(state == INVALID || (tag != aligned_access_offset && tag != GLOBAL_NULL)) {
-		load_cache_entry(aligned_access_offset, (startIndex%cachesize));
-#if DUAL_LOAD == 1
-		prefetch_cache_entry((aligned_access_offset+CACHELINE*pagesize), ((startIndex+CACHELINE)%cachesize));
-#endif
+		load_cache_entry(aligned_access_offset);
 		pthread_mutex_unlock(&cachemutex);
 		double t2 = MPI_Wtime();
 		stats.loadtime+=t2-t1;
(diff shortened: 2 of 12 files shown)


kostis:
Can we also have some test that shows some measurable performance improvements from this PR?

lundgren87:
> Can we also have some test that shows some measurable performance improvements from this PR?

### Example 1
Synthetic throughput benchmark ([Source code](https://github.com/lundgren87/paratest/tree/prefetch)) of loading pages from a remote node, ensuring that all data is allocated on and initialized by Node 0. All other nodes fetch all pages sequentially from Node 0. Run for 2, 3 and 5 nodes (with 1, 2 and 4 reading nodes as depicted below). The number of pages loaded per remote access is listed on top of each bar.
bw_prefetch
DISCLAIMER: This shows *only* the page loading throughput and nothing else. It is not expected to scale for multiple nodes as additional nodes perform exactly the same work as each node beyond Node 0, and instead congestion should (eventually) slow the system down for additional nodes.

### Example 2
DAPXY benchmark ([Source code](https://github.com/lundgren87/paratest)) on total 2GB data (2x1GB arrays) with data allocated across all nodes. Loading primarily occurs in the initialization phase (green bars).
**Two ArgoDSM nodes**
twonodes

**Four ArgoDSM nodes**
fournodes
The initialization time (green sections) is significantly reduced with prefetching enabled. During execution, data is already expected to be present on the nodes (some exceptions may occur due ArgoDSM allocation issues) as each node initializes the data it works on, hence prefetching has no real impact on the blue sections.

src/backend/mpi/swdsm.cpp
```diff
@@ -541,291 +529,251 @@ void write_back_writebuffer() {
...
-	if(tmptag == lineAddr && tmpstate != INVALID){
+	/* Return if cache entry is already up to date. */
+	if(cacheControl[start_index].tag == line_addr &&
+			cacheControl[start_index].state != INVALID){
 		sem_post(&ibsem);
 		return;
 	}
 
+	/* Adjust endidx to ensure the whole chunk to fetch is on the same node */
+	for(int j = start_index+CACHELINE, p = 1; static_cast<unsigned long>(j) <
```
kostis, line 564: Stylistic comment: wouldn't this read better in 3 (one per "part" of the for) instead of 2 lines?

kostis: Another comment related to this line (to `start_index` in particular), but it's a more general one, I noticed that the code uses `int` for some values (which should probably be `unsigned int`, really) in some places, and `std::size_t` in other places.

tests/prefetch.cpp
```diff
@@ -21,6 +21,12 @@ std::size_t size = 1<<30;
 namespace mem = argo::mempools;
 extern mem::global_memory_pool<>* default_global_mempool;
 
+/** @brief ArgoDSM page size */
```
kostis, line 24: Minor: no need for `/**` here; `/*` will do.
(Similar comment below.)

kostis (commented):
Looks OK overall.  The comments I've left in the code are minor.

The two more important comments I have are:

1. It would be nice for this PR to include a benchmark/program that shows the perf benefits this brings (and also ensure that these are not accidentally broken/undone in future updates).
2. Perhaps also include few tests that check that the most important errors we want to catch are indeed caught.

src/backend/mpi/swdsm.cpp
```diff
@@ -541,291 +529,251 @@ void write_back_writebuffer() {
...
-	unsigned long homenode;
-	unsigned long id = 1 << getID();
-	unsigned long invid = ~id;
 
-	if(loadtag>=size_of_all){//Trying to access/prefetch out of memory
-		return;
+	if(loadtag >= size_of_all){ // Address out of bounds
+		printf("Address %lu out of bounds.\n", loadtag);
+		exit(EXIT_FAILURE);
 	}
```
lundgren87, line 536: This should probably return instead of exit, as a segfault outside of the ArgoDSM memory range should not be handled by ArgoDSM. However, it is unclear if `load_cache_entry` can be called with an invalid "tag"/address in the first place.

lundgren87:
The PR has been updated to the latest master and prefetching now works for all allocation policies, verified on Jenkins. For all cyclic policies (naive, cyclic, prime-mapp and skew-mapp) the graphs in **Example 1** still apply, given that the `ARGO_ALLOCATION_BLOCK_SIZE` is at least as big as the `ARGO_LOAD_SIZE`, otherwise the load size is essentially capped to the allocation block size.

For first-touch allocation there is still scaling to observe (from 47mb/s without prefetching to 138mb/s with 64 page load size), but the fact that there is only one remote access to each page per node means that the cost of finding where the page has been first touched is significant compared to the local calculations in the cyclic policies.

The latest update also contains a number of tests that verify prefetching works as intended. After discussion with David I will look to add similar synthetic read, read-random, write and write-random throughput benchmarks to Jenkins in order to also test performance where, unlike our current benchmarks, remote reads and/or writes are not optimized away by specific data allocation patterns or initialization.

src/backend/mpi/swdsm.cpp
lundgren87: Both of these suggestions should be reflected in the latest commit.

src/backend/backend.hpp
```diff
@@ -213,7 +222,8 @@ namespace argo {
 			/**
 			 * @brief Backend internal type erased atomic store function
 			 * @param desired Pointer to the object that holds the desired value
-			 * @param size == sizeof(*desired)
+			 * @param count Number of element from desired to store
+			 * @param size size of each element in desired
```
kostis, line 226: Second `size` should read `Size`

kostis: Also, we should document the `count` param now.

src/backend/backend.hpp
```diff
@@ -222,11 +232,13 @@ namespace argo {
...
 			 */
 			void _store_public_owners_dir(const void* desired,
-				const std::size_t size, const std::size_t rank, const std::size_t disp);
+				const std::size_t size, const std::size_t count,
+				const std::size_t rank, const std::size_t disp);
 
 			/**
 			 * @brief Backend internal type erased atomic store function
 			 * @param desired The desired value to be stored
+			 * @param count Number of element from desired to store
```
kostis, line 241: element -> elements

src/backend/backend.hpp
```diff
@@ -264,7 +276,8 @@ namespace argo {
 			/**
 			 * @brief Backend internal type erased atomic load function
 			 * @param output_buffer Pointer to the memory location where the value of the object should be stored
-			 * @param size == sizeof(*output_buffer)
+			 * @param size Size of an element to load
+			 * @param count Number of element to load
```
kostis, line 280: element -> elements

src/backend/backend.hpp
```diff
@@ -273,11 +286,13 @@ namespace argo {
...
 			 */
 			void _load_public_owners_dir(void* output_buffer,
-				const std::size_t size, const std::size_t rank, const std::size_t disp);
+				const std::size_t size, const std::size_t count,
+				const std::size_t rank, const std::size_t disp);
 
 			/**
 			 * @brief Backend internal type erased atomic load function
 			 * @param output_buffer Pointer to the memory location where the value of the object should be stored
+			 * @param count Number of element to load
```
kostis, line 295: element -> elements

src/backend/mpi/swdsm.cpp
```diff
@@ -436,315 +433,262 @@ void handler(int sig, siginfo_t *si, void *unused){
...
+			/* Stop when either condition is not satisfied */
+			break;
+		}
+	}
 
-	void * lineptr = (char*)startAddr + lineAddr;
-
-	if(cacheControl[startidx].tag  != lineAddr){
-		if(cacheControl[startidx].tag  != lineAddr){
+	/* Allocate space for loading things */
```
kostis, line 508: Either make this comment more precise (there is no space allocation, it computes some sizes; what are the "things", etc.), or take it out.

src/backend/mpi/swdsm.cpp
```diff
@@ -436,315 +433,262 @@ void handler(int sig, siginfo_t *si, void *unused){
...
+	/* Temporarily store remotely fetched cache data */
+	std::vector<char> temp_data(fetch_size*pagesize);
+
+	/* Write back existing cache entries if needed */
+	for(std::size_t idx = start_index, p = 0; idx < end_index; idx+=CACHELINE, p+=CACHELINE){
+		/* Address and pointer to the data being loaded */
+		const std::size_t temp_addr = aligned_access_offset + p*block_size;
+
+		/* Skip updating pages that are already present and valid in the cache */
+		if(cacheControl[idx].tag  == temp_addr && cacheControl[idx].state != INVALID){
```
kostis, line 535: Extra space before `==`

src/backend/mpi/swdsm.cpp
```diff
@@ -436,315 +433,262 @@ void handler(int sig, siginfo_t *si, void *unused){
...
-
-
-	homenode = getHomenode(prefetchtag);
-	unsigned long cacheIndex = prefetchline;
-	if(cacheIndex >= cachesize){
-		printf("idx > size   cacheIndex:%ld cachesize:%ld\n",cacheIndex,cachesize);
-		return;
+	/* If this node is a new sharer of at least one of the pages */
+	if(new_sharer){
+		/* Register ourselves on all newly shared pages the loadnode directory */
```
kostis, line 597: Specify who "ourselves" is.  Also loadnode -> load_node.

src/backend/mpi/swdsm.cpp
```diff
@@ -1253,3 +1199,29 @@ size_t argo_get_global_size(){return size_of_all;}
...
+	std::size_t aligned_address = align_backwards(
+			addr-reinterpret_cast<std::size_t>(startAddr), CACHELINE*pagesize);
+	if(dd::is_first_touch_policy()){
+		homenode = peek_homenode(aligned_address);
+	}else{
+		homenode = get_homenode(aligned_address);
+	}
+	std::size_t cache_index = getCacheIndex(aligned_address);
+
+	bool is_local;
```
kostis, line 1214: All this code can be simplified to a one-liner.
```C
   // Return true for pages which are either local or already cached
   return ((homenode == getID) || (cacheControl[cache_index].tag == aligned_address &&
￼			           cacheControl[cache_index].state == VALID));
```

src/data_distribution/base_distribution.hpp
```diff
@@ -81,6 +93,19 @@ namespace argo {
 					return 0;
 				}
 
+				/**
+				 * @brief compute offset into the home node's share of the memory
+				 * @param ptr address to find offset of
```
kostis, line 98: Something is not OK in this sentence.

src/data_distribution/first_touch_distribution.hpp
```diff
@@ -69,14 +77,46 @@ namespace argo {
...
+				 * @param do_first_touch attempt to first touch if true
 				 */
-				static void update_dirs (const std::size_t& addr);
+				static void update_dirs (const std::size_t& addr, bool do_first_touch = true);
 				/** @brief protects the owners directory */
 				std::mutex owners_mutex;
 
 			public:
+				virtual node_id_t peek_homenode(char* const ptr) {
+					std::size_t page_info[3];
```
kostis, line 88: We should introduce symbolic constants (named macros) for these 3's that appear in this line and in some other ones below.

src/backend/mpi/swdsm.cpp
lundgren87: I pushed a more elaborate comment including the changes requested, please let me know if it is clear enough.

src/data_distribution/base_distribution.hpp
lundgren87: This is reworded in the latest commit, let me know if the new version is okay.

kostis (approved):

lundgren87:
The Doxygen documentation of the added constant `page_info_size` was missing the leading double asterisk in the latest commit, and Jenkins complained. This is now fixed, but unfortunately means I have to re-request a review @kostis .

kostis (dismissed):

src/backend/mpi/swdsm.cpp
```diff
@@ -436,315 +433,264 @@ void handler(int sig, siginfo_t *si, void *unused){
...
-		cacheControl[startidx].tag = lineAddr;
 	}
-	else{
-		mprotect(lineptr,pagesize*CACHELINE,PROT_READ);
-	}
-
-	touchedcache[startidx] = 1;
-	cacheControl[startidx].state = VALID;
-	cacheControl[startidx].dirty=CLEAN;
+	/* Ensure to free space occupied by tempData */
```
davidklaftenegger, line 689: `tempData` is not used anywhere else. What is this comment about?

src/backend/mpi/swdsm.cpp
```diff
@@ -1253,3 +1201,19 @@ size_t argo_get_global_size(){return size_of_all;}
...
+
+bool _is_cached(std::size_t addr) {
+	argo::node_id_t homenode;
+	std::size_t aligned_address = align_backwards(
+			addr-reinterpret_cast<std::size_t>(startAddr), CACHELINE*pagesize);
+	if(dd::is_first_touch_policy()){
+		homenode = peek_homenode(aligned_address);
+	}else{
+		homenode = get_homenode(aligned_address);
+	}
```
davidklaftenegger, line 1213: if we are already doing virtual lookups for `data_distribution` I do wonder why this branch is needed, but this is probably beyond this PR to fix.

src/data_distribution/base_distribution.hpp
davidklaftenegger: assuming kostis meant the parameter `ptr`, which is the address of which the offset is to be found, I think the comment is perfectly OK :)

src/data_distribution/first_touch_distribution.hpp
```diff
@@ -11,6 +11,9 @@
 
 #include "base_distribution.hpp"
 
+/** @brief The size used by each page in the owners_dir */
+const std::size_t page_info_size = 3;
```
davidklaftenegger, line 15: could we document here how the 3 is computed, or at least what kind of changes would also require this number to change?

src/data_distribution/first_touch_distribution.hpp
```diff
@@ -69,52 +80,116 @@ namespace argo {
...
+					const std::size_t owners_dir_window_index = page_info_size * (addr / granularity);
 
 					std::unique_lock<std::mutex> def_lock(owners_mutex, std::defer_lock);
 					def_lock.lock();
 					/* handle all the necessary directory actions for the global address */
 					update_dirs(addr);
 					/* spin in case the values other than `ownership` have not been reflected to the local window */
 					do {
-						argo::backend::atomic::_load_local_owners_dir(&offset, rank, owners_dir_window_index+1);
+						argo::backend::atomic::_load_local_owners_dir(&offset, 1, rank, owners_dir_window_index+1);
```
davidklaftenegger, line 192: magic number 1, applies to other uses of this function as well. I assume this means there is a directory stored on node 1, but I wonder why node 1 is magic and not node 0.

src/backend/mpi/swdsm.h
```diff
@@ -312,22 +310,45 @@ unsigned long getCacheIndex(unsigned long addr);
...
  * @return addr-(start address of local process part of global memory)
  */
-unsigned long getOffset(unsigned long addr, char cloc = 0);
+std::size_t get_offset(std::size_t addr);
+/**
+ * @brief Gets the offset of an address on the local nodes part of the global memory
+ * @param addr Address in the global address space
+ * @return addr-(start address of local process part of global memory)
+ * @note This version does not invoke a first-touch call and instead
+ * returns SIZE_MAX if an address has not been first-touched
```
davidklaftenegger, line 335: Is `SIZE_MAX` in the code clear enough, or should we have a constant (that is set to `SIZE_MAX`) and use that instead?

src/backend/mpi/swdsm.cpp
```diff
@@ -436,315 +433,264 @@ void handler(int sig, siginfo_t *si, void *unused){
...
-			if(1ul<<n==ownid) {
-				owner = n; //just get rank...
-				break;
+	for(std::size_t i = 0; i < fetch_size; i+=CACHELINE){
+		if(pages_to_load[i]){
+			/* Check local pyxis directory if we are sharer of the page */
+			local_sharers[i] = (globalSharers[classification_index_array[i]])&node_id_bit;
+			if(local_sharers[i] == 0){
+				sharer_bit_mask[i*2] = node_id_bit;
+				new_sharer = true; //At least one new sharer detected
```
davidklaftenegger, line 588: do I understand this correctly that even when we always load 8 pages, the coherence data is stored separately for each page? I'm not asking for this to change, just making sure I understand it correctly.

davidklaftenegger (changes requested):
Overall looking very good. I won't be able to fully dig through the algorithm itself today, so I have not checked it for correctness. The only real issue I see is the magic number `1`, which is very similar to the issue with magic number `3` that @kostis mentioned in his previous review.

src/backend/mpi/swdsm.cpp
```diff
@@ -436,315 +433,264 @@ void handler(int sig, siginfo_t *si, void *unused){
...
-				mprotect(tmpptr2,blocksize,PROT_NONE);
+				argo_write_buffer->erase(idx);
 			}
+			/* Ensure the writeback has finished */
+			for(int i = 0; i < numtasks; i++){
+				if(barwindowsused[i] == 1){
+					MPI_Win_unlock(i, globalDataWindow[i]);
+					barwindowsused[i] = 0;
+				}
+			} // TODO: move this out to avoid unlocking for every page?
```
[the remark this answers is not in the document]
lundgren87: I will remove the TODO. It may offer a small performance increase, but this can be revisited if/when we write a dedicated cache module.

src/backend/mpi/swdsm.cpp
```diff
@@ -92,18 +88,10 @@ sem_t ibsem;
 /*Loading and Prefetching*/
 /**
  * @brief load into cache helper function
- * @param tag aligned address to load into the cache
- * @param line cache entry index to use
+ * @param aligned_access_offset memory offset to load into the cache
+ * @pre aligned_access_offset must be aligned as CACHELINE*pagesize
```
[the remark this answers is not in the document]
lundgren87: Fixed.

src/backend/mpi/swdsm.cpp
```diff
@@ -436,315 +433,264 @@ void handler(int sig, siginfo_t *si, void *unused){
...
 
-
-				cacheControl[startidx].state = INVALID;
-				cacheControl[startidx].tag = lineAddr;
-				cacheControl[startidx].dirty=CLEAN;
-
-				vm::map_memory(lineptr, blocksize, pagesize*startidx, PROT_NONE);
-				mprotect(tmpptr2,blocksize,PROT_NONE);
-
+				/* Downgrade all relevant pages on the owner node (P>S) */
```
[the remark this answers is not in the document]
lundgren87: This now reads "from private to shared" instead of (P>S).

src/backend/mpi/swdsm.cpp
lundgren87: It was not supposed to be there at all, and no longer is!

src/backend/mpi/swdsm.cpp
lundgren87: We load _at most_ 8 pages, as there may be conditions that prevent sequential reads of `ARGO_LOAD_SIZE` pages.

The coherence data is stored separately for each page in the pyxis directory (`globalSharers`), with one `unsigned long` storing sharers and one storing writers.

lundgren87:
I've pushed an update that should include the fixes that @davidklaftenegger asked for. 

However, I've discovered a failing test for the first-touch allocation prefetching on OpenMPI 4.1.1 and GCC 10.2 that does not seem to be present on OpenMPI 4.1.0+GCC 9 or older. It is also not triggered when `ARGO_DEBUG=ON`.

This does not affect correctness, but prefetching appears to occasionally fetch some page(s) less than expected. I will investigate during the weekend.

src/backend/mpi/swdsm.h
lundgren87: The latest push should now detail that the return value is `::invalid_offset` for peek_offset (and `::invalid_node_id` for peek_homenode). These will link to the definitions in first_touch_allocation.hpp.

lundgren87:
The latest commit contains a fix to the failing test, which seem to have been triggered by recent changes in loop vectorization when compiling with `-O3`. In order for `PrefetchTest.AccessPrefetched` to be deterministic under first-touch allocation, a memory fence was inserted after each page access to ensure that these accesses can not be reordered by either compiler or the CPU.

kostis (approved):

src/data_distribution/first_touch_distribution.hpp
```diff
@@ -11,6 +11,27 @@
...
+ * in the MPI backend. Changes to this must be applied in both locations
+ * and should probably be centralized at some point.
+ */
+constexpr std::size_t page_info_size = single_entry_size * 3;
+
+/** @brief This constant is used to represent an invalid node id */
+constexpr argo::node_id_t invalid_node_id = -1;
+
+/** @brief This constant is used to represent an invalid offset */
+constexpr std::size_t invalid_offset = SIZE_MAX;
```
davidklaftenegger, line 33: both of these sound like they belong to either global_ptr.hpp or base_distribution.hpp. Is there a good reason to have them defined only for first_touch?

src/data_distribution/global_ptr.hpp
```diff
@@ -35,21 +35,21 @@ namespace argo {
 
 			public:
 				/** @brief construct nullptr */
-				global_ptr() : homenode(-1), local_offset(0) {}
+				global_ptr() : homenode(-1), local_offset(SIZE_MAX) {}
```
davidklaftenegger, line 38: These should probably be invalid_node_id and invalid_offset.

src/data_distribution/global_ptr.hpp
```diff
@@ -35,21 +35,21 @@ namespace argo {
...
-				global_ptr(T* ptr, const std::string& sel = "")
-						: homenode(-1), local_offset(0), access_ptr(ptr) {
-					if (!sel.compare("getHomenode")) {
-						homenode = policy()->homenode(reinterpret_cast<char*>(ptr));
-					} else if (!sel.compare("getOffset")) {
-						local_offset = policy()->local_offset(reinterpret_cast<char*>(ptr));
-					} else {
+				global_ptr(T* ptr, const bool compute_homenode = true,
+									const bool compute_offset = true)
+						: homenode(-1), local_offset(SIZE_MAX), access_ptr(ptr) {
```
davidklaftenegger, line 48: same as above

src/data_distribution/global_ptr.hpp
```diff
@@ -89,6 +89,32 @@ namespace argo {
 				 * @return home node id
 				 */
 				node_id_t node() {
+					// If homenode is not yet calculated we need to find it
+					if(homenode == -1) {
```
davidklaftenegger, line 93: invalid_node_id?

src/data_distribution/global_ptr.hpp
```diff
@@ -89,6 +89,32 @@ namespace argo {
...
+
+				/**
+				 * @brief return the home node of the value pointed to, or a
+				 * default value if the page has not yet been first-touched
+				 * under first-touch allocation.
+				 * @return home node id
+				 */
+				node_id_t peek_node() {
+					// If homenode is not yet calculated we need to find it
+					if(homenode == -1) {
```
davidklaftenegger, line 108: invalid_node_id?

src/data_distribution/global_ptr.hpp
```diff
@@ -97,6 +123,31 @@ namespace argo {
 				 * @return local offset
 				 */
 				std::size_t offset() {
+					if(local_offset == SIZE_MAX) {
```
davidklaftenegger, line 126: invalid_offset?

src/data_distribution/global_ptr.hpp
```diff
@@ -97,6 +123,31 @@ namespace argo {
...
+
+				/**
+				 * @brief return the offset on the home node's local memory share
+				 * or a default value if the page has not yet been first-touched
+				 * under first-touch allocation.
+				 * @return local offset
+				 */
+				node_id_t peek_offset() {
+					// If homenode is not yet calculated we need to find it
+					if(local_offset == SIZE_MAX) {
```
davidklaftenegger, line 141: invalid_offset?

tests/prefetch.cpp
```diff
@@ -14,16 +15,25 @@
...
 /** @brief ArgoDSM cache size */
 constexpr std::size_t cache_size = size/8;
 
+namespace env = argo::env;
+namespace dd = argo::data_distribution;
 namespace mem = argo::mempools;
 extern mem::global_memory_pool<>* default_global_mempool;
 
+/** @brief ArgoDSM page size */
+const std::size_t page_size = 4096;
```
davidklaftenegger, line 32: this is now the 7th or so place where we define this constant. Maybe we should open a ticket to track all of them for a code cleanup PR.

davidklaftenegger (commented):
Looks good, only some inconsistent use of constants.

tests/prefetch.cpp
davidklaftenegger: #49 already exists

davidklaftenegger (approved):
If tests still pass we can merge.

PR #40, opened 2020-06-20
Upgrade to use the latest googletest (1.11.0)

kostis:
While at it, change the installation instructions and two places in the
code to specify that the unzipped dir should be called `googletest'
instead of having its release version suffix.

PR #42, opened 2020-06-25
Add a travis script

kostis:


kostis:
Most likely, this is not working yet, but it needs to be merged because it appears that a repository is tested on Travis only _after_ it contains a .travis.yml file.  So please, somebody review and merge this so that I can test it further.

kostis:
Just in case it is not clear (or the latest commit message got ignored) this PR now works.

It should be reviewed in order to be extended and/or merged.

lundgren87:
It appears that everything is in order judging by the last run of the travis script. The `FIXME` part requires a solution to #40 eventually.

Should this script run with `-DARGO_DEBUG=ON` (alternatively as separate build and test) or is it sufficient that the release version is tested?

PR #43, opened 2020-09-21
Standardise rounding in argo_initialize

### LINKED ISSUE #24: standardise functions for rounding etc.
See the conversations in PR #23 for suggested places where rounding / ceiling could be factored into helper functions.

This could improve code quality and ease code review.

lundgren87:
This PR aims to resolve #24 and comments related to it in #23.
It implements and uses a forwards alignment function instead of replicating alignment in the code.

This PR also ensures that no extra page is unnecessarily added to the cache when the requested `cache_size` is already aligned. A potential bug with `cache_size<4097` causing an eviction loop is also prevented by ensuring that the cache always fits at least two `CACHELINE`.

Ideally there would be no need for both `cache_size` and `cachesize` to be confusingly used in the backend, but this is out of the scope of what this PR aims to do.

### DIFF
--- a/src/backend/mpi/swdsm.cpp
+++ b/src/backend/mpi/swdsm.cpp
@@ -901,16 +901,25 @@ void * argo_gmalloc(unsigned long size){
 	return ptrtmp;
 }
 
+/**
+ * @brief aligns an offset (into a memory region) to the beginning of its
+ * subsequent size block if it is not already aligned to a size block.
+ * @param offset the unaligned offset
+ * @param size the size of each block
+ * @return the aligned offset
+ */
+std::size_t align_forwards(std::size_t offset, std::size_t size){
+	return (offset == 0) ? offset : (1 + ((offset-1) / size))*size;
+}
+
 void argo_initialize(std::size_t argo_size, std::size_t cache_size){
 	int i;
 	unsigned long j;
 	initmpi();
-	unsigned long alignment = pagesize*CACHELINE*numtasks;
-	if((argo_size%alignment)>0){
-		argo_size += alignment - 1;
-		argo_size /= alignment;
-		argo_size *= alignment;
-	}
+
+	/** Standardise the ArgoDSM memory space */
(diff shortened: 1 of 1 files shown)


kostis (approved):

src/backend/mpi/swdsm.cpp
```diff
@@ -901,16 +901,25 @@ void * argo_gmalloc(unsigned long size){
...
 
+/**
+ * @brief aligns an offset (into a memory region) to the beginning of its
+ * subsequent size block if it is not already aligned to a size block.
+ * @param offset the unaligned offset
+ * @param size the size of each block
+ * @return the aligned offset
+ */
+std::size_t align_forwards(std::size_t offset, std::size_t size){
+	return (offset == 0) ? offset : (1 + ((offset-1) / size))*size;
```
davidklaftenegger, line 912: The special case for `0` feels somewhat odd. Is this just to avoid the case where -1 / n would fall to 0 instead of -1? Maybe returning `0` instead of `offset` would be more clear?

src/backend/mpi/swdsm.cpp
```diff
@@ -922,16 +931,13 @@ void argo_initialize(std::size_t argo_size, std::size_t cache_size){
...
-		cachesize += argo_size;
-	} else {
-		cachesize += cache_size;
-	}
-	cachesize += pagesize*CACHELINE;
+	/** Limit cache_size to at most argo_size */
+	cachesize = std::min(argo_size, cache_size);
+	/** Round the number of cache pages upwards */
+	cachesize = align_forwards(cachesize, pagesize*CACHELINE);
+	/** At least two pages are required to prevent endless eviction loops */
```
davidklaftenegger, line 938: not two pages, but two cachelines / pagelines / cache entries / cache entry slots?

davidklaftenegger (approved):
overall looking good. maths appear to be correct.

PR #45, opened 2021-01-25
Selective coherence and new write buffer

lundgren87:
This PR aims to add selective coherence functionality to ArgoDSM through `argo::backend::selective_acquire(void* addr, std::size_t size)` and `argo::backend::selective_release(void* addr, std::size_t size)`. Selective coherence performs invalidation and downgrade just like their regular counterparts but limited to a region confined by `[addr, addr+size)`, without performing self-invalidation or self-downgrade on the remainder of the cache. Used in the correct situations, so long as the data race free property of the code is maintained, this can provide significant performance gains by reducing the amount of remote reads and writes.

This PR also includes a dedicated write buffer storing cache indexes to be written back to their remote homenodes, implemented with the goal of providing a FIFO-style buffer with the option to delete any given element from the buffer. This is required in order to delete a cache index that may exist anywhere in the buffer when performing selective coherence on the corresponding cached page. The size of the write buffer can be set through the environment variable `ARGO_WRITE_BUFFER_SIZE` and defaults to 512 pages if not set. The number of least recently added pages written back once attempting to add a new page to a full write buffer can be set through the environment variable `ARGO_WRITE_BUFFER_WRITE_BACK_SIZE` and defaults to 32 pages.

### DIFF
--- a/src/backend/backend.hpp
+++ b/src/backend/backend.hpp
@@ -118,6 +118,68 @@ namespace argo {
 		 */
 		void release();
 
+
+		/**
+		 * The following selective coherence functions are implemented individually
+		 * by each backend
+		 */
+
+		/**
+		 * @brief backend internal type erased function for selective acquire
+		 * @param addr pointer to the start of an ArgoDSM memory region
+		 * @param size size of the region pointed to by addr
+		 * @sa    selective_acquire
+		 * @warning For internal use only - DO NOT USE UNLESS YOU KNOW WHAT YOU ARE DOING
+		 */
+		void _selective_acquire(void* addr, std::size_t size);
+
+		/**
+		 * @brief backend internal type erased function for selective release
+		 * @param addr pointer to the start of an ArgoDSM memory region
+		 * @param size size of the region pointed to by addr
+		 * @sa    selective_release
+		 * @warning For internal use only - DO NOT USE UNLESS YOU KNOW WHAT YOU ARE DOING
+		 */
+		void _selective_release(void* addr, std::size_t size);
+
+		/**
+		 * The following selective coherence functions are generic interfaces to the
+		 * actual backend implementations
+		 */
+
+		/**
+		 * @brief causes a node to self-invalidate the specified pages from its cache,
+		 *        and thus getting any updated values on subsequent accesses
+		 * @param addr pointer to the start of the memory region to self-invalidate
+		 * @param size the size of the memory region pointed to by addr. A size equal
+		 *        to zero results in that no pages are selectively acquired.
+		 * @note  Selective coherence operations do not uphold the data consistency
+		 *        semantics of regular coherence operations and should be used with
+		 *        this in mind. Specifically, write ordering with respect to writes
+		 *        outside of the selective coherence region is not upheld.
+		 */
+		template<typename T>
+		void selective_acquire(T* addr, std::size_t size) {
+			_selective_acquire(static_cast<void*>(addr), size);
+		}
+
+		/**
+		 * @brief causes a node to self-downgrade specified pages from its cache,
+		 *        making all previous writes to be propagated to the homenode,
+		 *        visible for nodes calling acquire (or other synchronization primitives)
+		 * @param addr the start of the memory region to self-invalidate
+		 * @param size the size of the memory region pointed to by addr. A size equal
+		 *        to zero results in that no pages are selectively released.
+		 * @note  Selective coherence operations do not uphold the data consistency
+		 *        semantics of regular coherence operations and should be used with
+		 *        this in mind. Specifically, write ordering with respect to writes
+		 *        outside of the selective coherence region is not upheld.
+		 */
+		template<typename T>
+		void selective_release(T* addr, std::size_t size) {
+			_selective_release(static_cast<void*>(addr), size);
+		}
+
 		namespace atomic {
 			using namespace argo::atomic;
 
--- a/src/backend/mpi/coherence.cpp
+++ b/src/backend/mpi/coherence.cpp
@@ -6,6 +6,7 @@
 
 #include "../backend.hpp"
 #include "swdsm.h"
+#include "write_buffer.hpp"
 #include "virtual_memory/virtual_memory.hpp"
 
 // EXTERNAL VARIABLES FROM BACKEND
@@ -61,6 +62,11 @@ extern argo_byte *touchedcache;
  * @brief workcomm is needed to poke the MPI system during one sided RMA
  */
 extern MPI_Comm workcomm;
+/**
+ * @brief Write buffer to ensure selectively handled pages can be removed
+ * @deprecated This should eventually be handled by a cache module
+ */
+extern write_buffer<std::size_t>* argo_write_buffer;
 
 namespace argo {
 	namespace backend {
@@ -81,7 +87,6 @@ namespace argo {
 			// Lock relevant mutexes. Start statistics timekeeping
 			double t1 = MPI_Wtime();
 			pthread_mutex_lock(&cachemutex);
-			pthread_mutex_lock(&wbmutex);
 			sem_wait(&ibsem);
 
 			// Iterate over all pages to selectively invalidate
@@ -94,10 +99,11 @@ namespace argo {
 				// If the page is dirty, downgrade it
 				if(cacheControl[cache_index].dirty == DIRTY){
 					mprotect((char*)start_address + page_address, block_size, PROT_READ);
-					cacheControl[cache_index].dirty = CLEAN;
 					for(int i = 0; i <CACHELINE; i++){
 						storepageDIFF(cache_index+i,page_address+page_size*i);
 					}
+					argo_write_buffer->erase(cache_index);
+					cacheControl[cache_index].dirty = CLEAN;
 				}
 
 				// Optimization to keep pages in cache if they do not
@@ -140,7 +146,6 @@ namespace argo {
 
 			// Release relevant mutexes
 			sem_post(&ibsem);
-			pthread_mutex_unlock(&wbmutex);
 			pthread_mutex_unlock(&cachemutex);
 		}
 
@@ -159,7 +164,6 @@ namespace argo {
 			// Lock relevant mutexes. Start statistics timekeeping
 			double t1 = MPI_Wtime();
 			pthread_mutex_lock(&cachemutex);
-			pthread_mutex_lock(&wbmutex);
 			sem_wait(&ibsem);
 
 			// Iterate over all pages to selectively downgrade
@@ -171,10 +175,11 @@ namespace argo {
 				// If the page is dirty, downgrade it
 				if(cacheControl[cache_index].dirty == DIRTY){
 					mprotect((char*)start_address + page_address, block_size, PROT_READ);
-					cacheControl[cache_index].dirty = CLEAN;
 					for(int i = 0; i <CACHELINE; i++){
 						storepageDIFF(cache_index+i,page_address+page_size*i);
 					}
+					argo_write_buffer->erase(cache_index);
+					cacheControl[cache_index].dirty = CLEAN;
 				}
 			}
 			// Make sure to sync writebacks
@@ -194,7 +199,6 @@ namespace argo {
 
 			// Release relevant mutexes
 			sem_post(&ibsem);
-			pthread_mutex_unlock(&wbmutex);
 			pthread_mutex_unlock(&cachemutex);
 		}
 	} //namespace backend
--- a/src/backend/mpi/coherence.cpp
+++ b/src/backend/mpi/coherence.cpp
@@ -1,0 +1,201 @@
+/**
+ * @file
+ * @brief This file implements selective coherence mechanisms for ArgoDSM
+ * @copyright Eta Scale AB. Licensed under the Eta Scale Open Source License. See the LICENSE file for details.
+ */
+
+#include "../backend.hpp"
+#include "swdsm.h"
+#include "virtual_memory/virtual_memory.hpp"
+
+// EXTERNAL VARIABLES FROM BACKEND
+/**
+ * @brief This is needed to access page information from the cache
+ * @deprecated Should be replaced with a cache API
+ */
+extern control_data *cacheControl;
+/**
+ * @brief globalSharers is needed to access and modify the pyxis directory
+ * @deprecated Should eventually be handled by a cache module
+ */
+extern unsigned long *globalSharers;
+/**
+ * @brief A cache mutex protects all operations on cacheControl
+ * @deprecated Should eventually be handled by a cache module
+ */
+extern pthread_mutex_t cachemutex;
+/**
+ * @brief ibsem is used to serialize all Infiniband (MPI) operations
+ * @deprecated Should not be needed once the cache module is implemented
+ */
+extern sem_t ibsem;
+/**
+ * @brief barwindowsused is needed to know which globalDataWindows
+ * to close.
+ * @deprecated No longer needed once proper API functions are in place
+ */
+extern char * barwindowsused;
+/**
+ * @brief globalDataWindows need to be explicitly closed by the
+ * caller of storepageDIFF
+ * @deprecated Should eventually be handled through API functions
+ */
+extern MPI_Win *globalDataWindow;
+/**
+ * @brief sharerWindow protects the pyxis directory
+ * @deprecated Should not be needed once the pyxis directory is
+ * managed from elsewhere through a cache module.
+ */
+extern MPI_Win sharerWindow;
+/**
+ * @brief Needed to update argo statistics
+ * @deprecated Should be replaced by API calls to a stats module
+ */
+extern argo_statistics stats;
+/**
+ * @brief Needed to update information about cache pages touched
+ * @deprecated Should eventually be handled by a cache module
+ */
+extern argo_byte *touchedcache;
+/**
+ * @brief workcomm is needed to poke the MPI system during one sided RMA
+ */
+extern MPI_Comm workcomm;
+
+namespace argo {
+	namespace backend {
+		void _selective_acquire(void *addr, std::size_t size){
+			if(size == 0){
+				// Nothing to invalidate
+				return;
+			}
+
+			const std::size_t block_size = page_size*CACHELINE;
+			const std::size_t start_address = reinterpret_cast<std::size_t>(argo::virtual_memory::start_address());
+			const std::size_t page_misalignment = reinterpret_cast<std::size_t>(addr)%block_size;
+			std::size_t argo_address =
+				((reinterpret_cast<std::size_t>(addr)-start_address)/block_size)*block_size;
+			const std::size_t node_id = argo::backend::node_id();
+			const std::size_t node_id_bit = 1 << node_id;
+
+			// Lock relevant mutexes. Start statistics timekeeping
+			double t1 = MPI_Wtime();
+			pthread_mutex_lock(&cachemutex);
+			pthread_mutex_lock(&wbmutex);
+			sem_wait(&ibsem);
+
+			// Iterate over all pages to selectively invalidate
+			for(std::size_t page_address = argo_address;
+					page_address < argo_address + page_misalignment + size;
+					page_address += block_size){
+				const std::size_t cache_index = getCacheIndex(page_address);
+				const std::size_t classification_index = get_classification_index(page_address);
+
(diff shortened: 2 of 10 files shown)


lundgren87:
Jenkins successfully passed all 3 builds (testing pipeline #705-707, #708 in progress with last Doxygen fixes). Showing 20-40% performance improvements on Matmul, small improvements on Conjugate Gradient and inconclusive small differences (in both directions) on the remaining benchmarks.

src/backend/mpi/coherence.cpp
```diff
@@ -0,0 +1,167 @@
...
+extern pthread_mutex_t wbmutex;
+extern sem_t ibsem;
+extern void * startAddr;
+extern char * barwindowsused;
+extern MPI_Win *globalDataWindow;
+extern MPI_Win sharerWindow;
+extern argo_statistics stats;
+extern argo_byte *touchedcache;
+extern MPI_Comm workcomm;
+extern WriteBuffer<unsigned long>* write_buffer;
```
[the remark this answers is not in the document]
davidklaftenegger: probably `std::size_t` in this particular instance.

src/backend/mpi/swdsm.cpp
```diff
@@ -50,16 +53,14 @@ char * pagecopy;
...
-/** @brief Most recent entry in the writebuffer*/
-unsigned long  writebufferstart;
-/** @brief Least recent entry in the writebuffer*/
-unsigned long  writebufferend;
+/** @brief A write buffer storing cache indexes */
+WriteBuffer<unsigned long>* write_buffer;
+/** @brief  Size of the write buffer*/
+size_t write_buffer_size;
+/** @brief  Number of write buffer elements to write back when full */
+size_t write_buffer_write_back_size;
```
davidklaftenegger, line 61: instead of being global variables, could these two be moved to the WriteBuffer class?

src/backend/mpi/swdsm.cpp
```diff
@@ -1287,6 +1278,6 @@ void printStatistics(){
 void *argo_get_global_base(){return startAddr;}
 size_t argo_get_global_size(){return size_of_all;}
 
-inline unsigned long get_classification_index(uint64_t addr){
+unsigned long get_classification_index(uint64_t addr){
```
davidklaftenegger, line 1281: should this be `static inline` instead? It might be required to move to the header file, and I don't know whether the static is needed then.

src/backend/mpi/write_buffer.hpp
```diff
@@ -0,0 +1,113 @@
...
+ * @tparam T the type of the write buffer
+ */
+template<typename T>
+class WriteBuffer
+{
+	private:
+		/**
+		 * @brief This container holds cache indexes that should be written back
+		 */
+		std::deque<T> _buffer;
```
davidklaftenegger, line 26: good enough for now, but do you have an intuition about using another kind of structure here?

tests/backend.cpp
```diff
@@ -382,6 +383,188 @@ TEST_F(backendTest, atomicFetchAddPointer) {
...
+ */
+TEST_F(backendTest, writeBufferLoad) {
+	const std::size_t array_size = 4000000; // Just under max size 16Mb
+	const std::size_t num_writes = array_size/20; // Not too many random writes
+	int* array = argo::conew_array<int>(array_size);
+
+	// Random device to ensure accesses are irregular in order to expose
+	// random ordering between writebacks to different nodes
+	std::random_device rd;
+	std::mt19937 rng(rd());
```
davidklaftenegger, line 534: Question for @kostis
Is it okay to have a unit test that produces a truly random sequence, or should the seed here be made a constant to make sure all encountered bugs are perfectly reproducible?

src/backend/mpi/coherence.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+/** @brief The size of an ArgoDSM page */
+static const std::size_t pagesize = 4096;
+
+namespace argo {
+	namespace backend {
+		void selective_acquire(void *addr, size_t size){
+			if(size == 0){
```
davidklaftenegger, line 32: maybe it should be spelled out in the documentation of the function that passing size zero results in a no-op?

src/backend/mpi/coherence.cpp
```diff
@@ -0,0 +1,167 @@
...
+				// Nothing to invalidate
+				return;
+			}
+
+			const std::size_t block_size = pagesize*CACHELINE;
+			const std::size_t page_padding = reinterpret_cast<unsigned long>(addr)%block_size;
```
davidklaftenegger, line 38: maybe `page_misalignment` captures it better than `page_padding`?

src/backend/mpi/swdsm.cpp
```diff
@@ -150,70 +151,62 @@ unsigned long isPowerOf2(unsigned long x){
   return retval;
 }
 
-void flushWriteBuffer(void){
-	unsigned long i,j;
-	double t1,t2;
-
-	t1 = MPI_Wtime();
+void flush_write_buffer(){
```
davidklaftenegger, line 154: Could this function be part of the WriteBuffer class?

src/backend/mpi/swdsm.cpp
```diff
@@ -150,70 +151,62 @@ unsigned long isPowerOf2(unsigned long x){
...
-	writebufferend = 0;
-
 	pthread_mutex_unlock(&wbmutex);
-	t2 = MPI_Wtime();
+	double t2 = MPI_Wtime();
 	stats.flushtime += t2-t1;
 }
 
-void addToWriteBuffer(unsigned long cacheIndex){
+void add_to_write_buffer(unsigned long cache_index){
```
davidklaftenegger, line 190: Shouldn't this be a member of the WriteBuffer class?

src/backend/mpi/swdsm.cpp
```diff
@@ -512,31 +505,29 @@ unsigned long getOffset(unsigned long addr){
 	return offset;
 }
 
-void write_back_writebuffer() {
-	unsigned long i;
-	unsigned long oldstart;
-	unsigned long idx,tag;
+void write_back_write_buffer() {
```
davidklaftenegger, line 508: Maybe this also can move to the WriteBuffer class?

src/backend/mpi/coherence.cpp
```diff
@@ -0,0 +1,167 @@
...
+extern pthread_mutex_t cachemutex;
+extern pthread_mutex_t wbmutex;
+extern sem_t ibsem;
+extern void * startAddr;
+extern char * barwindowsused;
+extern MPI_Win *globalDataWindow;
+extern MPI_Win sharerWindow;
+extern argo_statistics stats;
+extern argo_byte *touchedcache;
+extern MPI_Comm workcomm;
```
davidklaftenegger, line 23: Please document for all of these why they are needed. I think most/all of them could eventually be moved to proper API calls, but we need to write classes or APIs for them first. E.g. a cache class, a writebuffer class, and properly defined "backend" actions that deal with epochs in a defined way.

davidklaftenegger (changes requested):
Please add an `@note` to the comments of `selective_acquire` and `selective_release` explaining that they do not cause global acquire/release semantics even when used on the same address? I fear this might be confusing, as the C++11 memory model uses a similar interface but different semantics.

There is a lot of code replication from the backend, which would be nice to encapsulate in some unified functions. Maybe we need a cache class for this, but in any case fixing the replication is beyond the scope of this PR.

tests/backend.cpp
kostis: If the outcome of the test is deterministic, it is OK, I think.

src/backend/mpi/coherence.cpp
lundgren87: This is now reflected in the documentation in backend.hpp.

src/backend/mpi/swdsm.cpp
lundgren87: The issue with this is that `get_classification_index` requires access to `classificationSize` which is only a member of swdsm.cpp. swdsm.h is included in multiple translation units and defining a static `classificationSize` in swdsm.h does not work to solve this problem. For now I am leaving the inlining to the compiler, if you have a better solution feel free to share it otherwise Jenkins should show if inlining carries any importance.

src/backend/mpi/coherence.cpp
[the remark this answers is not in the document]
lundgren87: The writebuffer has been changed to use `std::size_t`.

src/backend/mpi/swdsm.cpp
lundgren87: Both variables are now part of the WriteBuffer class.

src/backend/mpi/write_buffer.hpp
lundgren87: Nothing in particular, open to suggestions and easy to change down the line.

src/backend/mpi/coherence.cpp
lundgren87: Changed as suggested.

src/backend/mpi/swdsm.cpp
lundgren87: `flush_write_buffer` is now `WriteBuffer::flush()`.
`write_back_write_buffer` is now `WriteBuffer::flush_partial()`
`add_to_write_buffer` is now `WriteBuffer::add(T val)`
`WriteBuffer::erase(T val)` is used to erase a single element.

src/backend/mpi/swdsm.cpp
lundgren87: flush_write_buffer is now WriteBuffer::flush().
write_back_write_buffer is now WriteBuffer::flush_partial()
add_to_write_buffer is now WriteBuffer::add(T val)
WriteBuffer::erase(T val) is used to erase a single element.

src/backend/mpi/swdsm.cpp
lundgren87: flush_write_buffer is now WriteBuffer::flush().
write_back_write_buffer is now WriteBuffer::flush_partial()
add_to_write_buffer is now WriteBuffer::add(T val)
WriteBuffer::erase(T val) is used to erase a single element.

src/backend/mpi/coherence.cpp
lundgren87: These are now documented, all with `@deprecated` tags as they should not be needed long-term and replaced by appropriate API functions or carried out in other modules.

tests/backend.cpp
lundgren87: I have not changed this test in the latest push. If this is not urgent perhaps it should be left as an issue (Ensure deterministic results in backend test).

lundgren87:
One additional minor change pushed, the `WriteBuffer` class is now `write_buffer` in order to comply with the ArgoDSM code policy.

lundgren87:
Jenkins passes all tests/benchmarks with the same performance as mentioned in the first comments. Final push simply fixes a missing doxygen documentation for the return value of `write_buffer::operator=`.

davidklaftenegger (commented):
This looks very nice, a properly abstracted write buffer cleans up quite a bit of code!

src/backend/mpi/swdsm.cpp
davidklaftenegger: very nice :)

src/backend/mpi/coherence.cpp
```diff
@@ -0,0 +1,208 @@
...
+namespace argo {
+	namespace backend {
+		void selective_acquire(void *addr, size_t size){
+			if(size == 0){
+				// Nothing to invalidate
+				return;
+			}
+
+			const std::size_t block_size = pagesize*CACHELINE;
+			const std::size_t start_address = reinterpret_cast<std::size_t>(argo::virtual_memory::start_address());
```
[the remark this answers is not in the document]
lundgren87: This may be worth thinking about. I believe that, in concept, it makes sense for the default representation of an "address" return type to be `void*`.

However, we use the returned pointer from `start_address()` primarily in pointer arithmetics, in which case we have traditionally cast it to `char*`. It also is used directly to `mprotect` in a couple of places.
In this patch I have since our type discussion used `std::size_t` to represent addresses, hence the cast to `std::size_t` instead of `char*`.

Which would be the correct signature in this case? Or is it simply more _safe_ to return a `void*` and let each use cast as appropriate if necessary?

src/backend/backend.hpp
```diff
@@ -118,6 +118,33 @@ namespace argo {
...
+		 *        and thus getting any updated values on subsequent accesses
+		 * @param addr	the start of the memory region to self-invalidate
+		 * @param size  the size of the memory region to self-invalidate. A size equal
+		 * to zero results in that no pages are selectively acquired.
+		 * @note  Selective coherence operations do not uphold the data consistency
+		 * semantics of regular coherence operations and should be used with this
+		 * in mind. Specifically, write ordering with respect to writes outside of
+		 * the selective coherence region is not upheld.
+		 */
+		void selective_acquire(void* addr, size_t size);
```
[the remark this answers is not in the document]
lundgren87: cppreference states that implicit conversions occur (and are allowed):
> A prvalue pointer to any (optionally cv-qualified) object type T can be converted to a prvalue pointer to (identically cv-qualified) void. The resulting pointer represents the same location in memory as the original pointer value. If the original pointer is a null pointer value, the result is a null pointer value of the destination type.

And that the following qualifies as a `prvalue`
> &a, the built-in address-of expression;

I therefore assume that the conversion from:
`argo::backend::selective_acquire(T& object, sizeof(T));`
to:
`argo::backend::selective_acquire(void* addr, size_t size);`
is both implicit and allowed.

The selective coherence operations make no assumption about the type of the address they are passed, and treat it solely as a memory address. 

The reason for this choice is originally just an old implementation relic, but I can at the same time not see a simple way (without multiple explicit instantiations) of providing separate templated definitions of `argo::backend::selective_acquire/release` for each backend. The definition of the templated function must be available at compile time, as the linker will otherwise not find a matching definition for each `T` without explicit instantiations if I understand things correctly. Am I missing a straightforward way to accomplish this?

src/backend/mpi/coherence.cpp
```diff
@@ -0,0 +1,208 @@
...
+			// Iterate over all pages to selectively invalidate
+			for(std::size_t page_address = argo_address;
+					page_address < argo_address + page_misalignment + size;
+					page_address += block_size){
+				const std::size_t cache_index = getCacheIndex(page_address);
+				const std::size_t classification_index = get_classification_index(page_address);
+
+				// If the page is dirty, downgrade it
+				if(cacheControl[cache_index].dirty == DIRTY){
+					argo_write_buffer->erase(cache_index);
```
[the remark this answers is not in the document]
lundgren87: It is currently safe placed both before and after the writeback as `ibsem` and `cachemutex` currently prevent simultaneous operations from happening on both the cache and the write buffer.

I also believe that for any foreseeable future, simultaneous cache operations on the same cached page will not be permitted (even in the parallel backend I lock access to at least each cache index separately), and therefore it should be safe to perform these in any order so long as there is mutual exclusion on the cache index.

However, as the cache is not set to `CLEAN` or `INVALID` until after `storepageDIFF` has been performed in `swdsm.cpp`, it may be preferable to keep this behaviour uniform and erase the write buffer entry after creating the diff if only to keep the order of operations the same. I will make this change in the next push.

src/backend/mpi/swdsm.cpp
```diff
@@ -760,8 +648,6 @@ void prefetch_cache_entry(unsigned long prefetchtag, unsigned long prefetchline)
 				mprotect(tmpptr2,blocksize,PROT_NONE);
 
 			}
-			pthread_mutex_unlock(&wbmutex);
```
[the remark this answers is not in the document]
lundgren87: `wbmutex` is only documented to protect access to the write buffer. In fact, I believe it was entirely surplus to requirements even for that purpose as every access to the write buffer was also protected by `cachemutex` _and_ `ibsem`. I therefore see this as "cleaning up on the go".

src/backend/backend.hpp
```diff
@@ -118,6 +118,33 @@ namespace argo {
...
 
+		/**
+		 * @brief causes a node to self-invalidate the specified pages from its cache,
+		 *        and thus getting any updated values on subsequent accesses
+		 * @param addr	the start of the memory region to self-invalidate
+		 * @param size  the size of the memory region to self-invalidate. A size equal
+		 * to zero results in that no pages are selectively acquired.
+		 * @note  Selective coherence operations do not uphold the data consistency
+		 * semantics of regular coherence operations and should be used with this
+		 * in mind. Specifically, write ordering with respect to writes outside of
```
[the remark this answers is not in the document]
lundgren87: Fixed in the latest push.

src/backend/backend.hpp
```diff
@@ -118,6 +118,33 @@ namespace argo {
...
+
+		/**
+		 * @brief causes a node to self-downgrade specified pages from its cache,
+		 *        making all previous writes to be propagated to the homenode,
+		 *        visible for nodes calling acquire (or other synchronization primitives)
+		 * @param addr	the start of the memory region to self-invalidate
+		 * @param size  the size of the memory region to self-invalidate. A size equal
+		 * to zero results in that no pages are selectively released.
+		 * @note  Selective coherence operations do not uphold the data consistency
+		 * semantics of regular coherence operations and should be used with this
```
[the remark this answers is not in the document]
lundgren87: Fixed in the latest push.

src/backend/backend.hpp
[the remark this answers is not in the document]
lundgren87: This is now changed according to the original suggestion:
`selective_acquire(T* addr, size_t size);` and
`selective_release(T* addr, size_t size);` respectively.

src/backend/mpi/coherence.cpp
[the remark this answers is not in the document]
lundgren87: Both setting the cache to `CLEAN` and removing the cache index from the write buffer now occurs _after_ the call to `storepageDIFF`.

src/backend/mpi/write_buffer.hpp
```diff
@@ -0,0 +1,307 @@
...
+extern MPI_Win* globalDataWindow;
+
+/**
+ * @brief		Tracking of which MPI RDMA windows are open
+ * @deprecated 	Prototype implementation, this should not be accessed directly
+ */
+extern char* barwindowsused;
+
+/** @brief Page size based on backend definition */
+const std::size_t page_size = GRAN;
```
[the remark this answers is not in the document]
lundgren87: As an intermediate solution, `page_size` is now defined in `src/backend/mpi/swdsm.h` and used in both selective coherence and the write buffer. I will log an issue upon this PR being merged to at some future point centralized this for all of ArgoDSM.

src/backend/mpi/write_buffer.hpp
```diff
@@ -0,0 +1,307 @@
...
+			_max_size(argo::env::write_buffer_size()/CACHELINE),
+			_write_back_size(argo::env::write_buffer_write_back_size()/CACHELINE) { }
+
+		/**
+		 * @brief	Copy constructor
+		 * @param	other The write_buffer object to copy from
+		 */
+		write_buffer(const write_buffer & other) {
+			// Ensure protection of data
+			std::unique_lock<std::mutex> lock_other(other._buffer_mutex);
```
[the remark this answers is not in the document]
lundgren87: I was debating this when I wrote the code, and seeing as you also hold this opinion I changed the locks that do not rely on deferred locking to use `std::lock_guard`. In c++17 `std::lock_guard` is superceded by `std::scoped_lock`, just as a note in case we would migrate at some point.

src/backend/mpi/swdsm.h
```diff
@@ -121,6 +121,13 @@ static const argo_byte WRITER=4;
 /** @brief Constant for reader states */
 static const argo_byte READER=5;
 
+/**
+ * @brief The size of a hardware memory page
+ * @note  This should be better centralized for all
+ *        modules and backen implementations
```
davidklaftenegger, line 127: *backend

lundgren87: Fixed in latest push.

lundgren87:
The latest push contains the typo fix mentioned by David, along with two minor function signature fixes picked up by Doxygen.

All Jenkins benchmarks and tests pass with the same result in terms of performance as above.

davidklaftenegger (approved):

PR #46, opened 2021-02-03
refactor global_tas_lock

davidklaftenegger:
By introducing a typedef for the internal field type we will be able to
change the implementation of the lock without breaking the code that
uses the global_tas_lock.


lundgren87 (approved):
This looks fine and passes Jenkins tests/benchmarks.

The only minor question I have is whether `field_type` would suffice rather than `internal_field_type`, simply for ease of use. Disregard if it could in any way be unclear which _"field"_ is being referred to.

PR #47, opened 2021-02-12
Fix compilation with -DARGO_DEBUG=ON

kostis:
Pass a CXX compiler option for bypassing a known issue
(https://github.com/open-mpi/ompi/issues/5157) in the C++
bindings for OpenMPI, which are deprecated anyway.

While at it, fix two small things I've noticed.


davidklaftenegger (approved):
looks okay, jenkins passes, can be merged.

PR #48, opened 2021-02-17
improved global TAS lock algorithm

davidklaftenegger:
The global TAS lock can elide a global synchronization iff the previous holder of the lock resided on the same ArgoDSM node. In this scenario, a local synchronization operation is sufficient.

Using cohort locks will still be better when the locks are contended, but maybe this is helpful for uncontended locks.


lundgren87 (approved):
This looks fine to me, David and I have discussed the comment on L54-72 and come to the conclusion that this should be the correct implementation given SC for DRF. Jenkins also passes.

PR #51, opened 2021-02-22
Upgrade to cmake > 3.3

kostis:
Newer versions of cmake require that a project specifies its version
numbers as part of the project() calls.

cmake versions > 3.12 simply issue a warning when this is not followed
but it is expected that at some point in the future this warning will
turn into an error.  So, this PR makes this future-compatible.


davidklaftenegger (approved):
looks harmless and jenkins does not complain

PR #54, opened 2021-03-13
Improve first-touch data distribution

IoanAnev:
This PR improves the _first-touch_ data distribution so that it passes ownership of a page to a remote node, when its backing store is not of sufficient size to host more pages. Briefly, the node first checks if it has sufficient backing store size, and if not, then right-wise checks for availability in the rest of the nodes.

This PR also:
* Enables back pre-fetching for the _first-touch_ data distribution.
* Decentralizes the ownership CAS operation done in `first_touch()`.
* Performs the MPI RMA operations to the public windows in bulk.
* Limits these operations to a specific node, rather than broadcasting.
* Detaches the directory actions from `homenode()` and moves them to `dir_actions()`.
* Reverts the tests in `backend.cpp` and `uninitialized.cpp`, which were changed due to _first-touch_ in #44.

Note: The function `dir_actions()` is invoked both in the `homenode()` and `local_offset()` functions, which eliminates the pre-condition of the `getHomenode()` function to have to be invoked before `getOffset()` for a specific address.

### DIFF
--- a/src/data_distribution/first_touch_distribution.hpp
+++ b/src/data_distribution/first_touch_distribution.hpp
@@ -8,7 +8,6 @@
 #define argo_first_touch_distribution_hpp argo_first_touch_distribution_hpp
 
 #include <mutex>
-#include <iostream>
 
 #include "base_distribution.hpp"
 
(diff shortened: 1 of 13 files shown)


IoanAnev:
Jenkins tests pass for the _naive_ and _first-touch_ distribution. Of course, we might want to check if there is no issue with the benchmarks having an equal to the data structures initialization size. Performance with this patch seems to degrade for the FFT benchmark, while the results for the rest of the benchmarks are the same.

src/data_distribution/first_touch_distribution.hpp
```diff
@@ -41,107 +45,122 @@ namespace argo {
...
-				node_id_t n;
-				for(n = 0; n < base_distribution<instance>::nodes; n++) {
-					if (n != argo::backend::node_id()) {
-						argo::backend::atomic::_store_public_dir(&id, sizeof(std::size_t), n, owner_window_index);
-						argo::backend::atomic::_store_public_dir(&argo::backend::backing_offset(), sizeof(std::size_t), n, owner_window_index+1);
+				for(n = 0; n < static_cast<std::size_t>(base_distribution<instance>::nodes); n++) {
+					if (n != rank) {
+						argo::backend::atomic::_store_public_owners_dir(&homenode, sizeof(std::size_t), n, owners_dir_window_index);
+						argo::backend::atomic::_store_public_owners_dir(&offset, sizeof(std::size_t), n, owners_dir_window_index+1);
+						argo::backend::atomic::_store_public_owners_dir(&rank, sizeof(std::size_t), n, owners_dir_window_index+2);
```
lundgren87, line 160: The one real concern I have with this approach is in this code block. This is essentially a broadcast of 3 values to _every_ node in the system, which will cost `3*(n-1)` remote operations for when each page is initially touched. With four nodes this may seem trivial, but with 64 nodes this is 189 remote operations the first time _each_ page is touched.

Each remote operation consists of the following sequence:
```
MPI_Win_lock(MPI_LOCK_EXCLUSIVE, ...);
MPI_Put(desired, 1, t_type, rank, disp, 1, t_type, owners_dir_window); // 1 value
MPI_Win_unlock(...)
```
The `Put` operation is guaranteed to have completed when `MPI_Win_unlock` returns, so there is no optimization to be had when placing three of them in succession. If instead, we write back all 3 values at once, since they are simply 3 successive integers, we reduce the amount of remote operations from `3*(n-1)` to simply `n-1`, as such:
```
/* put all 3 values in a buffer */
MPI_Win_lock(MPI_WIN_EXCLUSIVE, ...);
MPI_Put(buffer, 3, t_type, rank, disp, 3, t_type, owners_dir_window); // all 3 values
MPI_Win_unlock(...);
```
This still leaves expensive broadcasting in the system. I believe it would be significantly cheaper to write this to _one_ location, the so called `cas_node` you are already using for this page, and have each node query the `cas_node` the first time they load a page and only then add it to the local table. This would cost _at most_ the same number of remote operations as the modified broadcasting solution (one per node not performing the first touch = `n-1` remote operations), but they will be deferred to the point where the node actually wants the data. The worst-case cost is still the same, but this assumes that _every_ node touches _every_ page, and the average cost should be _significantly_ lower.

I suspect that this approach will overall perform much better in a first-touch scenario where data is not shared by multiple nodes, and that the cost of the current solution is hidden behind initialization that is not included in time measurements in our benchmarks.

lundgren87 (commented):
I think the solution to move the allocation to the next available node when our backing store is full is good, but I have some concerns on the increasing broadcasting cost for each allocation which is currently `3*(n-1)` remote operations per allocation for `n` nodes. I have proposed an alternative solution in my inline comment which greatly reduces the amount of remote operations, but it should probably be discussed before deciding on which approach to take.

I think we should also make sure to include initialization in some benchmark, as I do not believe we can treat initialization of data as "insignificant", especially in applications where initialization is not strictly necessary, when data is written and read only during execution of the main benchmark code.

lundgren87 (approved):
With the alterations according to the comments above in place, and some small optimizations, this should now be ready to merge.

PR #55, opened 2021-04-19
Add option to initialize ArgoDSM with multi-threaded MPI support

lundgren87:
This PR adds the CMake option `ARGO_ENABLE_MT` that ensures ArgoDSM is initalized with thread level `MPI_THREAD_MULTIPLE`. The default value of this option is `OFF`, as ArgoDSM itself does not yet make use of multithreaded MPI communication. Setting this to `ON` is necessary in order to support use of the ArgoDSM library in applications that already rely on multi-threaded MPI operations.

I believe that a CMake option is a better choice than an environment variable, even if this requires rebuilding ArgoDSM to change the state, as this ensures that the feature remains in the same state (ON or OFF) even if there are changes to the environment.

### DIFF
--- a/src/backend/mpi/CMakeLists.txt
+++ b/src/backend/mpi/CMakeLists.txt
@@ -7,3 +7,11 @@ install(TARGETS argobackend-mpi
 	RUNTIME DESTINATION bin
 	LIBRARY DESTINATION lib
 	ARCHIVE DESTINATION lib)
+
+option(ARGO_ENABLE_MT
+	"Enable support for multi-threaded MPI in ArgoDSM (MPI_THREAD_MULTIPLE)" OFF)
+if(ARGO_ENABLE_MT)
+	target_compile_definitions(argobackend-mpi PRIVATE ARGO_ENABLE_MT=1)
+else()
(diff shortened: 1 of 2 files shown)


lundgren87:
ArgoDSM currently serializes all outgoing MPI communication in accordance with `MPI_THREAD_SERIALIZED`. The downside of defaulting to ON while ArgoDSM does _not_ use the multi-threaded features is that we restrict the amount of MPI versions and components that can be used, as `MPI_THREAD_MULTIPLE` has been notoriously buggy in combination with one-sided communication (RMA/RDMA).

I am all for parallelizing the MPI backend, but until we do so I believe it is safer to default to `MPI_THREAD_SERIALIZED` instead of restricting the selection of supported MPI versions and their modules/components.

lundgren87:
Some changes suggested by David implemented, original functionality preserved. Both the CMake option and the macro are now called `ARGO_ENABLE_MT`, and the macro is now always defined as either 1 or 0 depending on the CMake option setting.

davidklaftenegger (approved):

kostis (approved):

PR #56, opened 2021-04-22
Add API functions needed by the nanos6 runtime

lundgren87:
This PR adds API functionality needed by the nanos6 runtime in order to make more intelligent scheduling decisions.
`bool argo::is_argo_address(T* addr) `- Check if an address is in ArgoDSM memory
`int argo::get_homenode(T* addr) `- Get the homenode of an address in ArgoDSM memory
`std::size_t argo::get_block_size()` - Get the size in bytes of a memory block that is guaranteed to be contiguously backed on one single node

A C-interface for each function (not templated) also included, as well as unit tests checking the correctness of each function.


src/argo.h
```diff
@@ -47,5 +47,25 @@ int  argo_node_id();
  * @return The total number of ArgoDSM nodes
  */
 int  argo_number_of_nodes();
+/**
```
davidklaftenegger, line 50: consistency with the rest of the file would suggest to have an empty line here (applies multiple times in this file)

src/argo.h
```diff
@@ -47,5 +47,25 @@ int  argo_node_id();
...
+ * @brief Check if addr belongs in the ArgoDSM memory space
+ * @param addr A memory address
+ * @return True if addr is in the ArgoDSM memory space, else false
+ */
+bool is_argo_address(void* addr);
```
davidklaftenegger, line 55: C-interface functions should be prefixed with `argo_`: `argo_is_argo_address`

src/argo.h
```diff
@@ -47,5 +47,25 @@ int  argo_node_id();
...
+/**
+ * @brief Get the ArgoDSM node id where addr is physically backed
+ * @param addr A valid address in the ArgoDSM memory space
+ * @return The id of the node physically backing addr, or
+ * argo::data_distribution::invalid_node_id if it has not yet
+ * been first-touched under the first-touch allocation policy
+ * @pre addr must be an address in ArgoDSM memory
+ */
+int get_homenode(void* addr);
```
davidklaftenegger, line 64: `argo_get_homenode`

src/argo.h
```diff
@@ -47,5 +47,25 @@ int  argo_node_id();
...
+/**
+ * @brief Get the block size of the current allocation policy
+ * @return The block size of the current allocation policy in bytes
+ */
+size_t get_block_size();
```
davidklaftenegger, line 69: `argo_get_block_size()`

tests/api.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+	/* Touch an equal number of pages per node */
+	int first_writer = (argo::get_homenode(start)+1) % num_nodes;
+	int my_start = (node_id-first_writer+num_nodes)%num_nodes;
+	for(std::size_t s=page_size*my_start;
+			s<alloc_size-1;
+			s+=page_size*num_nodes) {
+		tmp[s] = c_const;
+	}
+	argo::barrier();
```
davidklaftenegger, line 90: does this imply that prefetching must not cause a homenode to be set if the prefetched page is still unused?

src/backend/mpi/mpi.cpp
```diff
@@ -180,6 +180,9 @@ namespace argo {
 			return _is_cached(reinterpret_cast<std::size_t>(addr));
 		}
 
+		std::size_t block_size() {
+			return data_distribution::is_cyclic_policy() ? env::allocation_block_size()*page_size : CACHELINE*page_size;
```
davidklaftenegger, line 184: I am still wondering why this is not instead `data_distribution::block_size` instead of the `is_cyclic_policy()` check here. Is this re-inventing virtual functions, or is there some other reason to use a check here? Is this only about the `CACHELINE` constant? No change requested, just wondering.

src/argo.hpp
```diff
@@ -69,6 +70,41 @@ namespace argo {
...
+	 * @return The id of the node physically backing addr, or
+	 * argo::data_distribution::invalid_node_id if it has not yet
+	 * been first-touched under the first-touch allocation policy
+	 * @pre addr must be an address in ArgoDSM memory
+	 */
+	template<typename T>
+	int get_homenode(T* addr) {
+		data_distribution::global_ptr<T> gptr(addr, false, false);
+		return data_distribution::is_first_touch_policy() ?
+			gptr.peek_node() : gptr.node();
```
davidklaftenegger, line 100: should `peek_node` not always be "correct" here? my understanding is that `node` causes a homenode to be forced to exist while `peek_node` can return an invalid one iff it is not yet set. As for non-first-touch there is always a homenode set, `peek_node` should return the same value?

davidklaftenegger (changes requested):
The function names in `extern "C"` space should be changed, the rest looks okay. I have some questions, but they probably relate to some other software architecture decisions/changes.

src/argo.hpp
lundgren87: I will rewrite the `peek_homenode` implementation of each policy that directly calculates homenodes to call `homenode`, and alter this accordingly.

src/backend/mpi/mpi.cpp
lundgren87: `data_distribution::block_size` is initialized to `env::block_size()` in every case, which defaults to 16. This is not true for the _first-touch_ or _naive_ page placement policies, which do not have any defined block sizes. The requirement for this function is to return the minimum block size that is contiguous in memory.

- For _first-touch_, that is always going to be one block. (debatable whether this should be `page_size` or `CACHELINE*page_size`, since first-touch does not take `CACHELINE` into account)
- For the naïve page placement policy, this is either `CACHELINE*block_size` or `backend::global_size()/argo::number_of_nodes`.

tests/api.cpp
lundgren87: Prefetching causes a homenode to be set, whether this is correct behavior or not can be discussed. This is just to ensure that each page is owned by a specific node regardless of prefetching (since it can be disabled through envvars).

lundgren87: Actually, this is not true, mistake on my end! Pages that do not have a homenode are not prefetched at all unless they are the target of the cache miss.

davidklaftenegger: then I do not understand how this allows to "Test that the number of pages owned by each node is equal" below. wouldn't prefetching change how many pages are owned per node?

lundgren87: Probably a bad explanation on my end.

Loading a page `p` that does not have an owner claims ownership of `p` on behalf of the loading node, since we know that we want to access this page in some way.
However, loading of the subsequent `ARGO_LOAD_SIZE-1` pages is subject to whether the pages are
1. owned by the owner of `p` (same homenode)
2. contiguous in backing store (`offset of p+1 == offset of p + page_size`)

If one of these does not hold true, prefetching of this and subsequent pages stops. A page that does not have an owner fails the first condition and prefetching stops.

src/backend/mpi/mpi.cpp
lundgren87: `data_distribution::policy_block_size()` now returns the appropriate block size for the chosen allocation policy. The interface remains the same, but the backend is not directly involved.

src/argo.hpp
lundgren87: Should be fixed now.

lundgren87:
The comments left by David should now be addressed, testing on Jenkins before re-requesting a review.

tests/api.cpp
davidklaftenegger: >2) is a condition that only ever fails in the first-touch policy.

if I fetch the final page in one of node0's cyclic allocation blocks, does it prefetch into node1, or does it fetch a block aligned to the end of node0's block, or does it fail the condition and stop prefetching?

lundgren87: Under the current design, it fails condition 1 (homenode) on the following page and stops prefetching.

lundgren87:
Jenkins passed builds with naive, cyclic and first-touch policies so correctness should not have changed (as expected). Re-requesting review.

src/data_distribution/global_ptr.hpp
```diff
@@ -101,19 +101,15 @@ namespace argo {
 				 * @brief return the home node of the value pointed to, or a
 				 * default value if the page has not yet been first-touched
 				 * under first-touch allocation.
-				 * @return home node id
+				 * @return home node id, or argo::data_distribution::invalid_node_id
+				 * if access_ptr has not been first-touched yet
```
davidklaftenegger, line 105: This comment is slightly incorrect on non-first-touch policies, because it will not return `invalid_node_id`

src/data_distribution/global_ptr.hpp
```diff
@@ -101,19 +101,15 @@ namespace argo {
...
-						if(is_first_touch_policy()) {
-							homenode = policy()->peek_homenode(
-									reinterpret_cast<char*>(access_ptr));
-						} else {
-							homenode = policy()->homenode(
-									reinterpret_cast<char*>(access_ptr));
-						}
+						// Avoid invoking first-touch
+						homenode = policy()->peek_homenode(
+								reinterpret_cast<char*>(access_ptr));
```
davidklaftenegger, line 112: should this avoid the write to `homenode` iff there is no change, or is that premature optimization?

src/data_distribution/global_ptr.hpp
```diff
@@ -134,19 +130,15 @@ namespace argo {
...
-						if(is_first_touch_policy()) {
-							local_offset = policy()->peek_local_offset(
-									reinterpret_cast<char*>(access_ptr));
-						} else {
-							local_offset = policy()->local_offset(
-									reinterpret_cast<char*>(access_ptr));
-						}
+						// Avoid invoking first-touch
+						local_offset = policy()->peek_local_offset(
+								reinterpret_cast<char*>(access_ptr));
```
davidklaftenegger, line 141: same question w.r.t. premature optimization as above.

src/data_distribution/global_ptr.hpp
```diff
@@ -134,19 +130,15 @@ namespace argo {
 				 * @brief return the offset on the home node's local memory share
 				 * or a default value if the page has not yet been first-touched
 				 * under first-touch allocation.
-				 * @return local offset
+				 * @return local offset, or argo::data_distribution::invalid_offset
+				 * if access_ptr has not been first-touched yet
```
davidklaftenegger, line 134: same comment about correctness as above.

davidklaftenegger (approved):
overall looks good, some minor comments about comments should be addressed or mentioned in an issue.

lundgren87:
The minor changes suggested by @davidklaftenegger have been applied.

src/data_distribution/global_ptr.hpp
lundgren87: This now mentions that the alternative return value only concerns first-touch allocation. Same concept applied below.

src/data_distribution/global_ptr.hpp
lundgren87: Since I corrected the comment above I applied changes along these lines as well.

davidklaftenegger (approved):
No new comments, still good.

kostis (approved):

kostis:
I was just reviewing the changes when this was merged...
Changes looked good anyway!

lundgren87:
David approved it Friday but I never got around to merging during the weekend. If you find anything, point it out and I'll open issues for it (or revert the PR if it's significant) @kostis

PR #59, opened 2021-07-09
Avoid first-touching pages by default in C++ alloc

### LINKED ISSUE #58: C++ allocators cause first-touch on first page with or without initialization
The C++ allocator interface under first-touch allocation erroneously first-touches the first page of each allocation even if `argo::allocation::no_initialize` is specified.

https://github.com/etascale/argodsm/blob/70e7392a6e5bb6fc7076f93f7e52f6471de83867/src/allocators/collective_allocator.hpp#L218-L224

This occurs in a few locations and should be fixed (by someone, probably myself, but logging this issue so I don't forget). The fix is most likely to change the creation of global ptrs from:
`global_ptr<void> gptr(ptr);`
to:
`global_ptr<void> gptr(ptr, false, false);`
in order to prevent homenode and offset from being calculated (read first-touched under first-touch allocation) unless initialization is desired.

lundgren87:
This PR fixes #58 by ensuring that the C++ (de)allocators and do not erroneously first-touch pages that are not set to be (de)initialized under the first-touch allocation policy. The cause of this error was that the default `global_ptr` constructor calculates the homenode and node local offset of the given address, which for _first-touch_ corresponds to first-touching the page.

https://github.com/etascale/argodsm/blob/70e7392a6e5bb6fc7076f93f7e52f6471de83867/src/data_distribution/global_ptr.hpp#L40-L55

By passing `false` to both `compute_homenode` and `compute_offset`, _first-touch_ of the page is delayed until after evaluation of the `initialize` parameter, and is only performed if this evaluates to true.

Jenkins testing run 408 and 409 show unchanged performance as expected, since this error only concerns the first page of each allocation.

### DIFF
--- a/src/allocators/collective_allocator.hpp
+++ b/src/allocators/collective_allocator.hpp
@@ -106,7 +106,7 @@ namespace argo {
 
 		void* ptr = collective_alloc(sizeof(T));
 		using namespace data_distribution;
-		global_ptr<void> gptr(ptr);
+		global_ptr<void> gptr(ptr, false, false);
 		// The home node of ptr handles initialization
 		if (initialize && argo::backend::node_id() == gptr.node()) {
 			new (ptr) T(std::forward<Ps>(ps)...);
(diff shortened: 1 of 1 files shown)


kostis (approved):
Looks good (of course).

I still think it's a good idea in general to be accompanying these PRs with some program that shows their benefit, but perhaps in this case the effort would have been disproportional.

It can be merged.

PR #60, opened 2021-10-01
Parse environment variables of any data type

IoanAnev:
This PR templates the `parse_env` function and makes it generic so
that we can parse environment variables of any data type.

This is needed for the NVM version of ArgoDSM when the nvm path
of the nvm device needs to be given from an environment variable.

### DIFF
--- a/src/env/env.cpp
+++ b/src/env/env.cpp
@@ -6,6 +6,7 @@
 
 #include <cstdlib>
 #include <iostream>
+#include <sstream>
 #include <stdexcept>
 #include <string>
 #include <utility>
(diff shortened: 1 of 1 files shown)


src/env/env.cpp
```diff
@@ -154,16 +155,21 @@ namespace {
...
+	template<typename T>
+	std::pair<bool, T> parse_env(std::string name, T fallback) {
+		auto env_get = std::getenv(name.c_str());
 		try {
-			if(env_value != nullptr) {
-				return std::make_pair(true, std::stoul(env_value));
+			if(env_get != nullptr) {
+				std::string env_string(env_get);
+				std::stringstream env_stream(env_string);
+				T env_value; env_stream >> env_value;
```
davidklaftenegger, line 171: two lines.

davidklaftenegger (approved):
If jenkins agrees with this code I see no reason not to merge this.

IoanAnev:
> If jenkins agrees with this code I see no reason not to merge this.

Hello David,
Everything passed on Jenkins.

PR #66, opened 2021-12-10
Fix erroneous write misses

lundgren87:
This PR adds the following:

- A fix for the scenario when two or more read misses occur simultaneously causing all but the first read miss to **wrongfully register as write misses** (which also triggers unnecessary local and _remote_ pyxis downgrades and empty page writebacks). The cause of this error was that we detect write misses as any miss on a page that is already present and valid in the cache, which also applies for simultaneous read misses that wait for the cachemutex while another thread fetches the page in question. By checking the origin of the miss (read or write), we can at least filter out these erroneous misses for x86-64 hardware for now.
- A commit fixing a missing `atomic` include detected by GCC11.
- Definition of `NDEBUG` for release (non-debug) builds, as it may be important to test that the assumption of page fault error bits and the `REG_ERR` register holds true for future builds without impacting performance for release builds. _Please dispute this commit if you do not feel it is a good solution._


lundgren87:
I will queue this for testing on Jenkins at some point later today.

kostis (dismissed):

lundgren87:
> Did the Jenkings runs pass?
> 
> @davidklaftenegger any comments on this before I merge it?

It's running on Jenkins now, I did not get time to set it up before the weekend. Will update with results when it's completed later today.

src/backend/mpi/swdsm.cpp
```diff
@@ -236,15 +236,28 @@ inline std::size_t align_backwards(std::size_t offset, std::size_t size) {
...
+#ifdef REG_ERR
+	/* On x86, get and decode the error number from the context */
+	const ucontext_t* ctx = static_cast<ucontext_t*>(context);
+	auto err_num = ctx->uc_mcontext.gregs[REG_ERR];
+	assert(err_num & X86_PF_USER);	//Assert signal from user space
+	assert(err_num < X86_PF_RSVD); 	//Assert signal is from read or write access
+	/* This could be further decoded by using X86_PF_PROT to detect
+	 * whether the fault originated from no page found (0) or from
+	 * a protection fault (1), but is not needed for this purpose. */
+	bool write_miss = err_num & X86_PF_WRITE; //True for write, false for read
```
davidklaftenegger, line 258: would the later code be easier if we set this variable outside the `#ifdef`, but replicate the "old" behaviour (first=read, second=write) in its value?

src/synchronization/global_tas_lock.hpp
```diff
@@ -11,6 +11,7 @@
 #include "../data_distribution/global_ptr.hpp"
 #include <chrono>
 #include <thread>
+#include <atomic>
```
davidklaftenegger, line 14: our coding style doesn't say anything about it, but should these includes be in alphabetic order?

src/backend/mpi/swdsm.cpp
```diff
@@ -236,15 +236,28 @@ inline std::size_t align_backwards(std::size_t offset, std::size_t size) {
 	return (offset / size) * size;
 }
 
-void handler(int sig, siginfo_t *si, void *unused){
+void handler(int sig, siginfo_t *si, void *context){
 	UNUSED_PARAM(sig);
-	UNUSED_PARAM(unused);
```
davidklaftenegger, line 241: does the compiler not complain about the unused context when REG_ERR is not defined?

davidklaftenegger (changes requested):
The algorithmic change looks very good.

I am worried about the debug build failing if REG_ERR is undefined, can you test this and make sure the build doesn't break in this scenario?

All requested changes could be dismissed by a comment why they are not needed.

src/backend/mpi/swdsm.cpp
lundgren87: I have re-written the code slightly according to this suggestion. Write misses are now classified as `read`, `write` or `undefined` with the latter as default, and we make optimizations in the case where `read` or `write` miss knowledge can be used without checking for the definition of `REG_ERR`.

src/backend/mpi/swdsm.cpp
lundgren87: I believe this should now be fixed.

lundgren87:
Updated with slight re-write in accordance with (I hope) David's thoughts. Tests passed on Jason.

lundgren87:
Simplified the logic to avoid false write misses per David's request, currently running on Jenkins.

lundgren87:
Jenkins passed and if no one has any further comments everything should be ready.

davidklaftenegger (approved):
looks good.

PR #67, opened 2021-12-22
Fix literal type in bitwise left shift operations

IoanAnev:
This PR casts the occurences of literal `1` where we bitwise left
shift it to the appropriate type (unsigned long, size_t, uintptr_t).
This fixes correctness when running up to more than 32 nodes.

### DIFF
--- a/src/backend/mpi/coherence.cpp
+++ b/src/backend/mpi/coherence.cpp
@@ -82,7 +82,7 @@ namespace argo {
 			std::size_t argo_address =
 				((reinterpret_cast<std::size_t>(addr)-start_address)/block_size)*block_size;
 			const node_id_t node_id = argo::backend::node_id();
-			const std::size_t node_id_bit = 1 << node_id;
+			const std::size_t node_id_bit = static_cast<std::size_t>(1) << node_id;
 
 			// Lock relevant mutexes. Start statistics timekeeping
 			double t1 = MPI_Wtime();
--- a/src/backend/mpi/swdsm.cpp
+++ b/src/backend/mpi/swdsm.cpp
@@ -285,7 +285,7 @@ void handler(int sig, siginfo_t *si, void *context){
 		offset = get_offset(aligned_access_offset);
 	}
 
-	unsigned long id = 1 << getID();
+	unsigned long id = static_cast<unsigned long>(1) << getID();
 	unsigned long invid = ~id;
 
 	pthread_mutex_lock(&cachemutex);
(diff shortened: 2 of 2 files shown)


IoanAnev:
I will be testing on Jenkins right away!

kostis:
To my taste, writing `static_cast<unsigned long>(1)` instead of `1UL` seems quite verbose, but I will let others more proficient in C++ than me to decide (@davidklaftenegger ?)

On a more general comment, I was under the impression that this PR would be more of the "let's eliminate `unsigned long` occurrences and change them to use proper types and/or more descriptive type names" kind, but it seems it does something else.

IoanAnev:
> To my taste, writing `static_cast<unsigned long>(1)` instead of `1UL` seems quite verbose, but I will let others more proficient in C++ than me to decide (@davidklaftenegger ?)
> 
> On a more general comment, I was under the impression that this PR would be more of the "let's eliminate `unsigned long` occurrences and change them to use proper types and/or more descriptive type names" kind, but it seems it does something else.

As per our discussion with @lundgren87 and @davidklaftenegger , we deemed that it is better for this PR to exclusively focus on fixing the issue and leave the replacement of `unsigned long` for another one.

As far as the literal type is concerned, we decided that we should use `static_cast` instead `ul`, as this is more of a C++ way of doing it.

davidklaftenegger (approved):
there is three different problems at play here

1) an immediate correctness problem that hits every 32nd node, starting at the 33rd, by losing status information due to 32bit shift operations being performed. (I am not sure whether nodes 34-64 are truly unaffected, but it does not really matter - this is broken and must be fixed ASAP).

2) inconsistent types for id bit handling. Using 1ul would have added to this, as 1ul is unsigned long only, but we use other types (and unsigned long is worse than std::size_t, but uint64_t would probably be even better. For neither of these there is a literal notation.)

3) The id handling limits us to 64 nodes, which can be fixed by switching away from fixed-width integers to either `std::vector<bool>` or some wrapper class.


This PR addresses only issue 1), and issue 3) makes fixing issue 2) obsolete. I suggest an issue be submitted for issue 3).

lundgren87:
Just a minor comment, but I agree with David what we probably should do is to ensure that these values are in fact _always_ 64-bit unsigned, for which `std::uint64_t` makes the most sense. `std::size_t` is only _guaranteed_ to be at least 16-bit, even though it in practice mostly can be assumed to be 64-bit. Static cast to this type also appears the most appropriate in C++.

I will log the issue for 3) tomorrow unless someone does this before I have time to.

PR #70, opened 2022-01-31
Replace unsigned long with appropriate types

### LINKED ISSUE #69: Replace unsigned long with appropriate types
The MPI backend uses `unsigned long` as primary integer type. These should be replaced by more descriptive/suitable types such as `std::size_t` for sizes and counters, `std::uintptr_t` for pointers and adresses and `std::uint64_t` for integers that are required to be 64 bits or larger.

lundgren87:
This PR fixes #69 by assigning more appropriate types in place of `unsigned long` in the MPI backend. The goal is for the types to be descriptive of their use case, ie using `std::size_t` for things related to size, `std::uint64_t` when 64-bit is required and `std::uintptr_t` for memory addresses.

A few lines have not been removed as they have no use in the code and I instead propose to remove them entirely in #71 .

### DIFF
--- a/src/backend/mpi/swdsm.cpp
+++ b/src/backend/mpi/swdsm.cpp
@@ -289,7 +289,7 @@ void handler(int sig, siginfo_t *si, void *context){
 			if(sharers != 0 && sharers != id && isPowerOf2(sharers)){
 				std::uint64_t ownid = sharers&invid;
 				argo::node_id_t owner = workrank;
-				for(argo::node_id_t n=0; n<numtasks; n++){
+				for(argo::node_id_t n = 0; n < numtasks; n++){
 					if((static_cast<std::uint64_t>(1)<<n)==ownid){
 						owner = n; //just get rank...
 						break;
@@ -327,7 +327,7 @@ void handler(int sig, siginfo_t *si, void *context){
 
 			/* remote single writer */
 			if(writers != id && writers != 0 && isPowerOf2(writers&invid)){
-				for(argo::node_id_t n=0; n<numtasks; n++){
+				for(argo::node_id_t n = 0; n < numtasks; n++){
 					if((static_cast<std::uint64_t>(1)<<n)==(writers&invid)){
 						owner = n; //just get rank...
 						break;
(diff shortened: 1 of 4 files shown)


src/backend/mpi/swdsm.cpp
```diff
@@ -78,7 +78,7 @@ int numtasks;
 /** @brief  rank/process ID in the MPI/ArgoDSM runtime*/
 int rank;
 /** @brief rank/process ID in the MPI/ArgoDSM runtime*/
-int workrank;
+argo::node_id_t workrank;
```
kostis, line 81: It seems strange to change the type of `workrank` and leave `rank` declared `int`.  Aren't these two related?

src/backend/mpi/swdsm.cpp
```diff
@@ -978,10 +970,10 @@ void argo_initialize(std::size_t argo_size, std::size_t cache_size){
 		memset(global_offsets_tbl, 0, offsets_tbl_size_bytes);
 	}
 
-	for(j=0; j<cachesize; j++){
-		cacheControl[j].tag = GLOBAL_NULL;
-		cacheControl[j].state = INVALID;
-		cacheControl[j].dirty = CLEAN;
+	for(std::size_t i=0; i<cachesize; i++){
```
kostis, line 973: Add spaces around `=` and `<`, for consistency with other `for` loops in this file.

kostis (changes requested):
Do the two small changes I suggested, and this can be merged, as far as I am concerned.

src/backend/mpi/swdsm.cpp
lundgren87: I've pushed another commit making sure all (hopefully) `for` loops use this standard instead of just fixing this one occurrence.

src/backend/mpi/swdsm.cpp
lundgren87: The only use of `rank` in the backend is to run `MPI_Comm_rank`. The whole way we initialize MPI should probably be re-written as I discovered there is more "dead code" related to this, and `rank` is one of the things that is never used in the code outside of the mentioned call.

kostis (approved):

davidklaftenegger (commented):
if testing is not complaining I don't see how this could introduce any issues, so no objections from my side.

kostis (approved):

PR #71, opened 2022-01-31
Clean up unused backend code

lundgren87:
This PR aims to remove code in the MPI backend that has not been used for a long time. Removing this code will reduce clutter and make the code easier to maintain and improve.

The code in question mainly concerns `argo_gmalloc`, which was replaced by the dedicated allocators in `src/allocators` long ago, as well as the lock related code that has been replaced by locks in `src/synchronization`.

### DIFF
--- a/src/backend/mpi/swdsm.cpp
+++ b/src/backend/mpi/swdsm.cpp
@@ -65,8 +65,6 @@ MPI_Group workgroup;
 MPI_Comm workcomm;
 /** @brief MPI window for communicating pyxis directory*/
 MPI_Win sharerWindow;
-/** @brief MPI window for communicating global locks*/
-MPI_Win lockWindow;
 /** @brief MPI windows for reading and writing data in global address space */
 MPI_Win *globalDataWindow;
 /** @brief MPI data structure for sending cache control data*/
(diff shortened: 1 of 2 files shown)


kostis (approved):

src/backend/mpi/swdsm.cpp
```diff
@@ -930,12 +876,8 @@ void argo_initialize(std::size_t argo_size, std::size_t cache_size){
 	tmpcache=globalSharers;
 	vm::map_memory(tmpcache, gwritersize, current_offset, PROT_READ|PROT_WRITE);
 
-	current_offset += gwritersize;
-	tmpcache=lockbuffer;
-	vm::map_memory(tmpcache, pagesize, current_offset, PROT_READ|PROT_WRITE);
-
 	if (dd::is_first_touch_policy()) {
-		current_offset += pagesize;
+		current_offset += gwritersize;
```
davidklaftenegger, line 880: this appears to be a change in behaviour. Is this a silent bugfix, or something that went wrong when refactoring?

lundgren87: > this appears to be a change in behaviour. Is this a silent bugfix, or something that went wrong when refactoring?

This change is prompted by the removal of `lockbuffer` of size `pagesize` from the memory mappings.

If I understand it correctly, `current_offset` is increased by the size of the _previous_ mapping ahead of every new mapping (in the sense of code formatting), which is why this is a little bit unclear. Since the mapping of `lockbuffer` (removed) no longer happens, I moved `current_offset += gwritersize` to the next position (replacing `current_offset += pagesize` for lockbuffer) to avoid changing behaviour. Perhaps changing the behaviour alltogether to increase after each mapping would be more clear.

Do let me know if I have misunderstood anything!

davidklaftenegger: sounds correct

davidklaftenegger (approved):
I see no problems in this pull request, it can be merged.

PR #72, opened 2022-02-01
Move tests to GitHub actions

kostis:
Travis is no longer a free service, so this PR moves running the tests
on GitHub actions instead. At some later point, we can add support for
testing on different configurations rather than Ubuntu latest and also
add caching of the various Argo dependencies.


lundgren87 (approved):
Assuming that we have access to two cores and not more (documentation indicates this, tests run 2 ArgoDSM nodes by default but 4 nodes catches more issues from my experience), as well as no conflicts going forward using googletest 1.10, this looks fine.

PR #74, opened 2022-02-03
Update Doxyfile.in to a more recent version

kostis:
Use a more recent `doxygen` version and automatically update the `Doxyfile.in` using `doxygen -u`.
The main reason for this is to avoid some warnings when using doxygen 1.8.0 (or more recent).


lundgren87 (approved):
See no problems.

PR #78, opened 2022-03-16
Add six coherence tests to backendTests

ioanev:
This PR introduces six tests which test **ArgoDSM coherence** by
writing to random indices on a set of pages (first set of tests) and
within a page (other set of tests).

**Random accesses within the range of several pages**:
- randAccessesBarrierArray
- randAccessesBulkySelectiveReleaseAcquireArray
- randAccessesPeriodicSelectiveReleaseAcquireArray

**Random accesses within a page**:
- randAccessesBarrierPage
- randAccessesBulkySelectiveReleaseAcquirePage
- randAccessesPeriodicSelectiveReleaseAcquirePage

**Coherence operations for each test type**:
- randAccessesBarrier*
  -- uses an Argo barrier.
- randAccessesBulkySelectiveReleaseAcquire*
  -- uses a bulky selective-release, an MPI barrier, and an acquire.
- randAccessesPeriodicSelectiveReleaseAcquire*
  -- uses periodic selective-release, an MPI barrier, and an acquire.

> NOTE: The tests use a constant seed in order to be deterministic.

### DIFF
--- a/tests/backend.cpp
+++ b/tests/backend.cpp
@@ -4,7 +4,10 @@
  * @copyright Eta Scale AB. Licensed under the Eta Scale Open Source License. See the LICENSE file for details.
  */
 
+#include <algorithm>
 #include <chrono>
+#include <mpi.h>
+#include <numeric>
 #include <random>
 
 #include "argo.hpp"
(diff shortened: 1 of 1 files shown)


kostis (changes requested):
This is a minor comment just to check whether by providing a comment the tests will be triggered for this PR.

First, I suggest you factor out the strange number `1048576` that various tests use into some global symbolic name (with a `#define`).

Second, `indice` is not a word.  It should read `index` or `indices`.

ioanev:
As far as the number `1048576` is concerned, it was simply chosen because execution was taking a little bit too much time for these tests to complete when using the number `2097152` that was already used in other tests. IMO the `array_size` should be test-specific.
What does @davidklaftenegger & @lundgren87 have to say about this?

Fixed the `indice` issue. I missed adding the `s` at the end and since I was copying these lines, it propagated to all of the tests :)

kostis:
> we should not dirty this file with tons of global symbolic names. 

I fail to see how using an appropriate `#define`, as I suggested, dirties the file it appears in or makes any global name.

ioanev:
Applied the requested change in a seperate commit for ease of viewing, and just in case we don't want to keep it.

kostis:
> Applied the requested change

Well, you did not.  I mentioned using a `#define`, you did something else, more modern/C++-friendly, but different than what I suggested nonetheless.  But the main problem is below.

> ... in a seperate commit for ease of viewing, and just in case we don't want to keep it.

I do not understand the numbers that this commit uses (`1048576` and then later down `::array_size / 4`).  Can you please double-check them and/or explain them?  In particular, the line that reads:

```c++
constexpr std::size_t array_size = 1048576; // 4MB (ints)
```


More generally, and this is what has triggered this request for a change from my part, I really do not see the point in having hard-coded strange numbers in programs, instead of writing some more understandable expression like `1 << 20` or the even better `4 * 1024 * 1024` for expressing something that is supposed to be 4MB (as the comment claims).

ioanev:
> Well, you did not. I mentioned using a #define, you did something else, more modern/C++-friendly

Since we use constant expressions for global variables, I thought it would be better to be coherent.

> I do not understand the numbers that this commit uses (1048576 and then later down ::array_size / 4)

Yes, it makes sense to set `::array_size` directly to the value the **randAccesses*** tests use. (:heavy_check_mark:)

> More generally, and this is what has triggered this request for a change from my part, I really do not see the point in having hard-coded strange numbers in programs, instead of writing some more understandable expression like 1 << 20 or the even better 4 * 1024 * 1024 for expressing something that is supposed to be 4MB (as the comment claims).

Applied the change here as well and went with the bitwise shift approach to be coherent with `size`. (:heavy_check_mark:)
The value of `array_size` came out purely from _trial and error_ for **backendTests** to execute within a reasonable time.

<hr style="border:2px solid gray"> </hr>

**NOTE**: There are other tests which have hardcoded values like this in their test body.

Let me know if there is anything else.

kostis (approved):

PR #80, opened 2022-03-24
Remove unnecessary memsets from `argo_initialize`

### LINKED ISSUE #21: unneeded memsets in initialization
commit d6db61519ef9f6c11de7c97aae43204722fc3571 in PR #14 removes memsets which are deemed unnecessary.
As that PR is not working properly, this commit should be lifted out of it, applied to the current version, and performance tested. The performance testing is important, so this should not happen before we have proper performance tests: Without forced initialization, runtime may actually be negative impacted, contrary to expectation. Or it might help because of less TLB cluttering.

ioanev:
Performance is not negatively affected by the removal of the memsets.
Performance tests were conducted both on Jason (through Jenkins) & Karolina.
The Jenkins pipeline is green.

Fixes #21.

### DIFF
--- a/src/backend/mpi/swdsm.cpp
+++ b/src/backend/mpi/swdsm.cpp
@@ -896,18 +896,6 @@ void argo_initialize(std::size_t argo_size, std::size_t cache_size){
 									 MPI_INFO_NULL, MPI_COMM_WORLD, &offsets_tbl_window);
 	}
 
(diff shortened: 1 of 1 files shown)


lundgren87 (approved):
I see no reason not to merge this as it appears there is no discernible difference in performance for initialized or uninitialized workloads as well as no difference in performance on Jenkins.

PR #83, opened 2022-04-11
Completely clean state for each test group

ioanev:
This PR introduces a backend wrapper for `argo_reset_coherence` and adds the call prior to `argo_reset` for each group of tests. This helps us expose bugs previously hidden due to the dirty state of the different directories in the backend.

**Questions**:
1. Should we propagate the `int n` parameter of the `argo_reset_coherence` call (as seen below) to the backend wrapper?
https://github.com/etascale/argodsm/blob/0540544363178ba3bc79f95b38a20da31e9b323d/src/backend/mpi/swdsm.h#L204
2. Should we change the name of `argo_reset` to something more representative?
3. Minor: Should we avoid code duplication under `init` and `reset_coherence` in **singlenode.cpp**
for the `global_owners_dir` initialization by putting the code shown below in a seperate function?
https://github.com/etascale/argodsm/blob/0540544363178ba3bc79f95b38a20da31e9b323d/src/backend/singlenode/singlenode.cpp#L89-L104

Let me know if there's anything to be changed.

Issue #20, opened 2019-06-20
The Argo atomics do not install the data in the local cache


SakalisC:
In the MPI backend, when an Argo atomic instruction reads or modifies some data, it does an RDMA operation to access the data. The Argo coherence is bypassed and the data are never installed in the local cache. One would expect the data to be installed in the cache after an atomic access.

Relevant issue: The `loadOne` test fails in the second `ASSERT_EQ` (`ASSERT_EQ(i_const, *_i);`) even through the `while` loops before it checks for the exact same condition. The only difference is that the `while` loop uses an atomic instead of a normal load.

davidklaftenegger:
It should be noted that the failure of this test case is another bug, but failing in this particular assertion uncovered that atomic accesses are not stored in the cache.

PR #84, opened 2022-04-11
Modify atomic tests to use atomic load

ioanev:
Modify the atomic tests in backendTests to use `atomic::load` rather than pure memory loads as per issue #20.
This as a standalone PR, but also when cherry-picking the commit from #83, passes all tests with both `DEBUG=ON/OFF`.

### DIFF
--- a/tests/backend.cpp
+++ b/tests/backend.cpp
@@ -68,18 +68,19 @@ class backendTest : public testing::Test, public ::testing::WithParamInterface<i
 TEST_F(backendTest, atomicXchgAll) {
 	global_char _c(argo::conew_<char>(0));
 	argo::backend::atomic::exchange(_c, c_const);
-	ASSERT_EQ(c_const, *_c);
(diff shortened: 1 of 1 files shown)


tests/backend.cpp
```diff
@@ -167,30 +170,27 @@ TEST_F(backendTest, loadOne) {
 		std::chrono::system_clock::now() + deadlock_threshold;
 	while(argo::backend::atomic::load(_c) != c_const)
 		ASSERT_LT(std::chrono::system_clock::now(), max_time);
-	ASSERT_EQ(c_const, *_c);
```
lundgren87, line 170: This is a slight change of behaviour, but ultimately I think it is sufficient to check that the while loop has finished in order to verify the removed `ASSERT_EQ`.

lundgren87 (approved):
Aside from the one comment I left (which can most likely be left unchanged) I don't have any objections. There are tests that fail using OpenUCX over Infiniband (atomic tests), but these are not related to the changes here and I can document this in a separate issue.

PR #85, opened 2022-04-11
Reserve the last page of global memory for internal use

### LINKED ISSUE #22: necessity of reserved memory in global memory pool
Currently at https://github.com/etascale/argodsm/blob/bdda067df87645bb683c8595d7515689db2f06ac/src/mempools/global_mempool.hpp#L52

The global memory pool unconditionally adds some memory "for internal use", but it is unclear what this memory is used for, and whether this is actually needed.

The notorious corner case would be a setup where ArgoDSM only requests to handle a single page, which is not currently tested for.

ioanev:
This PR aims to resolve #22.

This patch reserves the last instead of the first page of global memory to host the currently available amount of pool memory (offset) and its associated lock structure to atomically (and also globally) update it.
Applying this patch, globally allocated data will start from the virtual address 0x200000000000 instead of 0x200000001000.

NOTE: This PR will close #81.

### DIFF
--- a/src/mempools/global_mempool.hpp
+++ b/src/mempools/global_mempool.hpp
@@ -55,12 +55,14 @@ namespace argo {
 					using namespace data_distribution;
 					base_distribution<0>::set_memory_space(nodes, memory, max_size);
 
-					// Move back one page as the last page is left for internal use
+					// Reset maximum size to the full memory size minus the space reserved for internal use
 					max_size -= reserved;
+					// Attach memory pool offset to the start of the reserved space
 					offset = new (&memory[max_size]) ptrdiff_t;
 
 					using tas_lock = argo::globallock::global_tas_lock;
-					tas_lock::internal_field_type* field = new (&memory[max_size+sizeof(std::size_t)]) tas_lock::internal_field_type;
+					// Attach internal lock field sizeof(ptrdiff_t) bytes after the start of the reserved space
+					tas_lock::internal_field_type* field = new (&memory[max_size+sizeof(std::ptrdiff_t)]) tas_lock::internal_field_type;
 					global_tas_lock = new tas_lock(field);
 
 					// Node0 makes sure that offset points to Argo's starting address
--- a/src/mempools/global_mempool.hpp
+++ b/src/mempools/global_mempool.hpp
@@ -65,8 +65,9 @@ namespace argo {
 					tas_lock::internal_field_type* field = new (&memory[max_size+sizeof(std::ptrdiff_t)]) tas_lock::internal_field_type;
 					global_tas_lock = new tas_lock(field);
 
-					// Node0 makes sure that offset points to Argo's starting address
-					if(backend::node_id() == 0){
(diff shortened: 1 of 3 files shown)


tests/prefetch.cpp
```diff
@@ -115,16 +115,16 @@ TEST_F(PrefetchTest, AccessPrefetched) {
...
 		 * For the first touch policiy, at most
 		 * (alloc_size/4096)/num_nodes are guaranteed to be contiguious 
 		 * in both global memory and backing store.
 		 */
-		stride = (load_size < ((size/page_size)/num_nodes)) ?
-				load_size : (size/page_size)/num_nodes - 1;
-		start_page = stride-1; //First stride may already be fetched in init
+		stride = (load_size < ((alloc_size/page_size)/num_nodes)) ?
+				load_size : (alloc_size/page_size)/num_nodes - 1;
+		start_page = 0; //First stride may already be fetched in init
```
lundgren87, line 127: This comment should be removed as it's no longer relevant.

tests/prefetch.cpp
```diff
@@ -115,16 +115,16 @@ TEST_F(PrefetchTest, AccessPrefetched) {
 		 * to be contiguious in both global memory and backing store
 		 */
 		stride = std::min(block_size, load_size);
-		start_page = block_size-1; //first block may already be fetched in init
+		start_page = 0; //first block may already be fetched in init
```
lundgren87, line 118: This comment should be removed as it's no longer relevant.

src/mempools/global_mempool.hpp
```diff
@@ -51,19 +51,21 @@ namespace argo {
 					auto nodes = backend::number_of_nodes();
 					memory = backend::global_base();
 					max_size = backend::global_size();
-					offset = new (&memory[0]) ptrdiff_t;
 					/**@todo this initialization should move to tools::init() land */
 					using namespace data_distribution;
 					base_distribution<0>::set_memory_space(nodes, memory, max_size);
+
+					// Move back one page as the last page is left for internal use
+					max_size -= reserved;
```
lundgren87, line 59: Can this line not be merged with L53 to ensure that it is coherent with the `reset()` implementation?
The comments for both should maybe reflect what is "moved back".

lundgren87 (changes requested):
This passes Jenkins for all allocation policies besides first-touch. First-touch appears to work after merging #90, but this should be confirmed on Jenkins.

I left a few minor comments that can be addressed or dismissed if you have a reason to disagree.

src/mempools/global_mempool.hpp
ioanev: The line mentioned is moved further down due to:
https://github.com/etascale/argodsm/blob/756ebfac81eec4422811de1eaab1affeec07858e/src/mempools/global_mempool.hpp#L57

lundgren87 (approved):
Changes look good. Performance on Jenkins is unchanged.

PR #88, opened 2022-04-28
Disable ArgoDSM atomics of size other than 4/8 bytes

lundgren87:
This PR disables 1, 2 and 16 byte datatype support for ArgoDSM atomics and their related unit tests, and leaves only 4 and 8 byte support. This is due to lack of support in UCX for atomic MPI operations of size other than 4 and 8 bytes.

While 1 and 2 byte MPI atomics are supposed to _all back to active (two-sided) messaging_ in UCX, this does not always seem to happen for a number of MPI atomics (`MPI_Fetch_and_op` used to implement `argo::backend::atomic::exchange` most notably in our case), and is most likely not desirable for performance reasons in the first place.

My suggestion is to leave this commented out in the code for now, and add an issue to further research this and based on findings either implement a solution or entirely remove the commented code and document this.

### DIFF
--- a/src/backend/mpi/mpi.cpp
+++ b/src/backend/mpi/mpi.cpp
@@ -66,18 +66,21 @@ extern sem_t ibsem;
  *
  * @param size The size of the datatype to be returned
  * @return An MPI datatype with MPI_Type_size == size
+ * @todo Either remove the 1/2 byte case entirely, or return at least 4 bytes and
+ * enforce 4 byte alignment of any MPI operations using the returned type.
  */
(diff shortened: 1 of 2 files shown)


kostis (approved):
Looks good, but "mpi" should read "MPI" in the comment text(s) of `tests/backend.cpp`

lundgren87:
> Looks good, but "mpi" should read "MPI" in the comment text(s) of `tests/backend.cpp`

Fixed all (hopefully) 4 instances of this.

kostis (approved):

PR #90, opened 2022-05-09
Align all memory allocations

lundgren87:
This PR ensures that all memory allocations are aligned to at least the alignment of `std::max_align_t` (defined since C++11). This is achieved by ensuring that the size of all allocations is a multiple of `std::max_align_t`. Aligning the size (end) of each allocation rather than the start of each allocation ensures that a scalar type allocation that has been deallocated (and added to the freelist) can be reused by other scalar types.

The PR fixes UCX errors regarding misaligned atomic instructions in the ArgoDSM atomic interface (also used by the global TAS lock and cohort locks).

I have also included two tests that:
- test the alignment for the different allocators functions available
- test reallocating deallocated memory

### DIFF
--- a/src/allocators/generic_allocator.hpp
+++ b/src/allocators/generic_allocator.hpp
@@ -221,7 +221,7 @@ namespace argo {
 
 					// Align the requested memory size in chunks of the alignment specified
 					// by max_align_t to ensure that all allocations are well aligned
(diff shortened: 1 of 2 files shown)


kostis (approved):

kostis:
Looks OK - I will merge.

PR #91, opened 2022-05-22
Change write-buffer container type from deque to set

ioanev:
This PR changes the container type of the write-buffer to be `std::set`.

Studying about C++ containers recently, I deemed it to be more appropriate for the container type of the write-buffer to be an associative container rather than the sequential container `std::deque`.

I've chosen the ordered associative-container type `set` for the following reasons:

- The cache indices that the write-buffer container holds have to be unique (no duplicates).
  - Each key in `set` is unique. If a key (cache index value) is already present in the buffer, subsequent additions of the same value will have no effect on the buffer (no-op).
- The cache indices of the write-buffer have to be sorted by home node id in ascending order before been written back.
  - We can attach a custom comparator to the `set` container that will handle the sorting in the background and we won't have to worry about doing it manually through the use of generic algorithms.

> **Note**
> Aside from sorting the write-buffer elements by home node id in ascending order, the custom comparator has been written to further sort the elements based on the cache index values (in ascending order). This is done solely for the reason to make it easier to implement a possible future optimization in which we will group adjacent pages and write them back in batches.

**Example:** A sequential insertion of cache indices 0-9 under the cyclic allocation policy with an alloc. block size of 1 will look like:
_buffer[0] == 0 (homenode: 0)
_buffer[1] == 4 (homenode: 0)
_buffer[2] == 8 (homenode: 0)
_buffer[3] == 1 (homenode: 1)
_buffer[4] == 5 (homenode: 1)
_buffer[5] == 9 (homenode: 1)
_buffer[6] == 2 (homenode: 2)
_buffer[7] == 6 (homenode: 2)
_buffer[8] == 3 (homenode: 3)
_buffer[9] == 7 (homenode: 3)

> **Warning**
> This PR has not been tested on Jenkins.

Let me know if you have any better ideas regarding the container type!

### DIFF
--- a/src/backend/mpi/write_buffer.hpp
+++ b/src/backend/mpi/write_buffer.hpp
@@ -7,12 +7,12 @@
 #ifndef argo_write_buffer_hpp
 #define argo_write_buffer_hpp argo_write_buffer_hpp
 
-#include <deque>
-#include <iterator>
 #include <algorithm>
-#include <mutex>
 #include <atomic>
+#include <iterator>
 #include <mpi.h>
+#include <mutex>
+#include <set>
 
 #include "backend/backend.hpp"
 #include "env/env.hpp"
@@ -56,15 +56,28 @@ template<typename T>
 class write_buffer
 {
 	private:
-		/** @brief This container holds cache indexes that should be written back */
-		std::deque<T> _buffer;
+		/**
+		 * @brief Key comparison function to sort elements by home node id and cache index in
+		 *        ascending order
+		 */
+		static bool comp_key(const T& l, const T& r) {
+			return (get_homenode(l) == get_homenode(r))
+			     ? l < r
+			     : (get_homenode(l)  < get_homenode(r));
+		}
+
+		/** 
+		 * @brief This associative container holds cache indices that should be written back
+		 * @note  Elements in the container are unique
+		 */
+		std::set<T, decltype(comp_key)*> _buffer{comp_key};
 
 		/** @brief The maximum size of the write buffer */
 		std::size_t _max_size;
 
 		/**
 		 * @brief The number of elements to write back once attempting to
-		 * add an element to an already full write buffer.
+		 *        add an element to an already full write buffer.
 		 */
 		std::size_t _write_back_size;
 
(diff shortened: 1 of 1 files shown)


lundgren87:
While I have no attachment to `std::deque` and am happy to see this changed to the most efficient data type, reviewing this I would like to suggest an alternative solution which I believe will perform better for current ArgoDSM as well as significantly simplify the code.

The only places in the code that still keeps multiple MPI windows locked during write back is in `argo::backend::selective_acquire/release`. Moving window unlocking into `storepageDIFF` (which locks it, this makes perfect sense) simplifies the code and by our previous experiments also improves performance, as keeping multiple windows locked for an extended time can be very expensive. **This change removes the need for the write buffer to be sorted at any point**, which reduces the code complexity of the write buffer implementation to at most that of your suggested implementation regardless of container choice.
 
The complexity of the remaining required operations are equal for `deque` vs `set`:
`std::find` (_O(log n)_)+`emplace_back` (_O(1)_) and `emplace` (_O(log n)_)
`pop_front` (_O(1)_) and `extract(const_iterator)` (_amortized O(1)_)

A sorted buffer _could_ be detrimental to performance as write back would favor lower (or higher, depending on sorting order and point of removal) indexed pages, and these risk being repeatedly written and re-added to the write buffer causing write backs of small incremental changes instead of possibly bulkier changes in regular FIFO order.

I suggest that we instead implement the change to `storepageDIFF` (and clean it up in general to get rid of `barwindowsused` alltogether) as well as find the best performing buffer container without a sorting restriction, and leave the sorting for a future bulk writeback patch if this becomes feasible in terms of both correctness and performance. The first step of implementing bulk diffs for whole pages does not require a sorted write buffer.

Please feel free to disagree or dispute any of my claims!

ioanev:
What you are suggesting seems reasonable from my side.

We can have the change to an ordered associative container type be part of the bulky write-backs PR, which will follow after the suggested changes in `storepageDIFF`.

As far as the container type based on the current implementation is concerned, since there would be no sorting restriction, the first thing that comes to my mind as a best choice is an _unordered_ associative container, and more specifically, `unordered_set`.

PR #93, opened 2022-05-25
Simplify write buffer flush functionality

lundgren87:
This PR simplifies the write buffer flush functionality by no longer
requiring the buffer to be sorted prior to flush, following #92 .

The actual page write back is also broken out into its own helper
function in order to reduce code duplication between
`write_buffer::_flush` and `write_buffer::flush_partial`.

Merging this should only be considered if no clear loss of
performance is shown in testing.

### DIFF
--- a/src/backend/mpi/write_buffer.hpp
+++ b/src/backend/mpi/write_buffer.hpp
@@ -145,7 +145,7 @@ class write_buffer
 			// Write back the page
 			mprotect(page_ptr, block_size, PROT_READ);
 			cacheControl[cache_index].dirty = CLEAN;
-			for(int i=0; i < CACHELINE; i++){
+			for(std::size_t i = 0; i < CACHELINE; i++){
 				storepageDIFF(cache_index+i, page_size*i+page_address);
 			}
 		}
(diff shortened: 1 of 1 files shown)


lundgren87:
No negative performance implications were shown for either allocation policy by not sorting the buffer prior to flushing it, tested for all five allocation policies on Jenkins.

src/backend/mpi/write_buffer.hpp
```diff
@@ -132,52 +132,35 @@ class write_buffer
...
+		void write_back_index(std::size_t cache_index) {
+			assert(cacheControl[cache_index].dirty == DIRTY);
+			std::uintptr_t page_address = cacheControl[cache_index].tag;
+			void* page_ptr = static_cast<char*>(
+				argo::virtual_memory::start_address()) + page_address;
+
+			// Write back the page
+			mprotect(page_ptr, block_size, PROT_READ);
+			cacheControl[cache_index].dirty = CLEAN;
+			for(int i=0; i < CACHELINE; i++){
```
kostis, line 148: perhaps `std::size_t` instead of `int` here?

kostis (approved):

kostis (approved):

PR #94, opened 2022-05-26
Include Googletest as submodule

lundgren87:
This PR incluides [GoogleTest](https://github.com/google/googletest) as a git submodule of ArgoDSM, and makes slight improvements to the submodule handling and verbose feedback overall. The latest version is fetched and built automatically through CMake if `ARGO_TESTS` is enabled. This is in accordance with the [Abseil Live at Head philosophy](https://abseil.io/about/philosophy#upgrade-support) employed by GoogleTest.

GoogleTest is built and links with the tests like before, but the its libraries and headers are no longer installed as part of the ArgoDSM project, as this should not be necessary.

432e93c updates the documentation regarding the submodules in general and the googletest submodule in particular.
Finally 8e7caf6 disables the step that fetches the googletest archive as part of the GitHub actions workflow.


lundgren87:
Passes tests on Jenkins but please carefully review (and try to break things on your end) before considering accepting!

kostis (dismissed):
Changes look OK.  However, I am wondering whether the following code
```
                message(STATUS "Submodule update: extern/qd")
                execute_process(COMMAND ${GIT_EXECUTABLE} submodule update
                        --init --recursive --remote extern/qd
                        WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
                        RESULT_VARIABLE GIT_SUBMOD_RESULT)
                if(NOT GIT_SUBMOD_RESULT EQUAL "0")
                        message(FATAL_ERROR "git submodule update --init --recursive "
                        "--remote extern/qd failed with ${GIT_SUBMOD_RESULT}, "
                        "please checkout submodules")
                else()
                        message(STATUS "Submodule update: extern/qd - Success")
                endif()
```
can be turned into a function that takes `extern/qd` (or even just `qd`) as a parameter, and then we can use this function for the similar code below it for `extern/googletest`, thus avoiding the code duplication.

davidklaftenegger (commented):
I'm not entirely sure about the "extern" directory name, but I guess it is too minor a detail to waste time on discussing the best name of it.

Some comments inline.

Rest is looking good, it worked both in online and offline mode for my setups.

lundgren87:
I think `submodules` is an appropriate name. I'll implement that and the changes requested if no one disagrees.

kostis (approved):

PR #96, opened 2022-06-02
Remove checks for using obsolete compilers

kostis:
It is highly unlikely that Argo will be compiled with anything else
than relatively modern versions of GCC or Clang. Remove some checks
around code which was not included if compiling with an old version
of GCC or some compiler othen than modern GCC or Clang.


davidklaftenegger (approved):
we don't have the manpower to maintain support for older compilers, so we don't even know if it would work on gcc-4.

PR #100, opened 2022-06-08
Fix a new Doxygen warning

kostis:


lundgren87 (approved):
Don't see any issues with merging this.

PR #111, opened 2022-06-15
Fix two empty while loop bodies

kostis:
by using {} instead of a ; for their empty body.


lundgren87 (approved):
This is valid code either way but this change perhaps makes it more clear that the body is empty.

PR #115, opened 2022-06-16
Simplify MPI initialization

lundgren87:
This PR removes the redundant MPI groups that were used to
initialize ArgoDSM.

Additionally, there is no need for both `workrank` and `rank`, as well
as `getID()` and `argo_get_nid()` as they do the exact same thing. This
PR removes `rank` and `getID()`, and renames `argo_get_nid()` to the
more clear `argo_get_node_id()`.

### DIFF
--- a/src/backend/mpi/coherence.cpp
+++ b/src/backend/mpi/coherence.cpp
@@ -87,7 +87,7 @@ namespace argo {
 
 			// Poke the MPI system to force progress
 			int flag;
-			MPI_Iprobe(MPI_ANY_SOURCE, MPI_ANY_TAG, workcomm, &flag, MPI_STATUS_IGNORE);
+			MPI_Iprobe(MPI_ANY_SOURCE, MPI_ANY_TAG, MPI_COMM_WORLD, &flag, MPI_STATUS_IGNORE);
 
 			std::lock_guard<std::mutex> ssi_time_lock(stats.ssi_time_mutex);
 			stats.ssi_time += t2-t1;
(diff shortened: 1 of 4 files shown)


lundgren87:
I am testing this on Jenkins to verify that this PR does not impact performance.

lundgren87:
> I am testing this on Jenkins to verify that this PR does not impact performance.
No performance impact as expected.

kostis (dismissed):
Looks good.

lundgren87:
> The reasoning for not using COMM_WORLD was interaction with other MPI code, which back then we anticipated might happen. Is this no longer a concern?

I do not think it is a big concern in theory, as ArgoDSM is responsible for initializing MPI and I would expect any application/user code not responsible for initializing MPI to create a private communicator on a subset (which can be the full set) of all MPI processes. In practice, maybe your concern is valid.

Inspecting MPI documentation I came across the following statement in multiple locations.
>  In particular, no library routine should use MPI_COMM_WORLD as the communicator; instead, a duplicate of a user-specified communicator should always be used.

Even if ArgoDSM performs MPI communication without providing direct library routines for this, I believe that the safest solution in this case is to create a copy of `MPI_COMM_WORLD` (for example `argo_comm`) using `MPI_Comm_dup` rather than creating a new group, and using this communicator for all MPI communication inside ArgoDSM. This should safeguard against MPI conflicts with any other code running inside (or encapsulating) ArgoDSM with the addition of at most a few lines of code.

I can write this patch tomorrow if this sounds like a good solution.

davidklaftenegger (dismissed):
I believe this to be a management decision, and I believe we have discussed the technical merits of both available choices. I am fine with either the current patch, or the suggested addition of nice code that improves potential interoperability.

lundgren87:
6cca164 introduces `argo_comm` as a duplicate of `MPI_COMM_WORLD`, which should isolate all ArgoDSM communication from that of other applications/users. With this change, I do not believe there is any reason to ever use `MPI_COMM_WORLD` outside of ArgoDSM initialization, as `argo_comm` is guaranteed to include the full set of MPI processes.

kostis (dismissed):
Looks good to me, but perhaps @davidklaftenegger should also take a look before this is merged.

kostis (approved):

kostis:
By leaving this unmerged for long time, a fair amount of conflicts needed to be resolved.
To avoid this situation happening again, I will now merge this.
Perhaps we should re-test it on Jenkings, just in case, after it is merged.

PR #116, opened 2022-06-16
Fix const errors in virtual memory

lundgren87:
This PR fixes an error in `virtual_memory/memfd` where a pointer to
const data is passed to `mmap` causing a compiler error when
building with MEMFD support. We can not pass a _pointer to const
data_ to `mmap` as `mmap` will attempt to alter the protection. The
_pointer itself_ can however be const.

Alternatively, If we explicitly want `ARGO_START` and
`ARGO_END` to point to const data, we need to cast the const
away with an explicit `const_cast` instead before passing it to
mmap. This does not feel like an intuitive solution to me, but correct
me if I'm wrong!

It also applies c++ casts where c casts were
used in `virtual_memory/anonymous`.

### DIFF
--- a/src/virtual_memory/anonymous.cpp
+++ b/src/virtual_memory/anonymous.cpp
@@ -23,7 +23,7 @@
 namespace {
 	/* file constants */
 	/** @todo hardcoded start address */
-	const char* ARGO_START = (char*) 0x200000000000l;
+	char* const ARGO_START = reinterpret_cast<char*>(0x200000000000l);
 	/** @todo hardcoded size */
 	const ptrdiff_t ARGO_SIZE = 0x80000000000l;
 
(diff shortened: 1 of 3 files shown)


davidklaftenegger:
> To me, it makes sense having both the _top-level_ and _low-level_ const for `ARGO_START` and `ARGO_END`. That is:

Why does it make sense? While we will never modify the data through this particular pointer, the memory region (ArgoDSM's memory) is hardly immutable during runtime, so I currently don't see any strong argument for or against the char to be const. The pointer should be const, as the patch currently does.

davidklaftenegger:
> Looks good, but shouldn't similar changes also happen to `shm.cpp` ?

in my opinion: yes

lundgren87:
I've applied the changes to `virtual_memory/shm.cpp` as well, all three modes now build without errors.

kostis (approved):
Is OK and mergable now.

PR #117, opened 2022-06-16
Specify which VM model to enable via an = option

kostis:
Instead of employing `-DARGO_VM_???` options for VM management
(and have complicated code to check that exactly one of them is enabled).

Besides simplifying the code, this will enable testing all options more
easily on GitHub actions.

For the moment, the same default has been kept, but it will need to be
revised when #110 is appropriately fixed.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -27,8 +27,9 @@ endforeach(src)
 
 
 unset(vm_libs)
-
+set(ARGO_VM "SHM" CACHE STRING "Select how to handle virtual addresses")
 # Exactly one of these must be enabled (or use the default)
+set_property(CACHE ARGO_VM PROPERTY STRINGS ANONYMOUS MEMFD SHM)
 if(ARGO_VM STREQUAL "MEMFD")
 # "Handle virtual addresses using an anonymous memory file. Requires kernel 3.17+."
 	set(vm_sources memfd.cpp)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -26,36 +26,20 @@ foreach(src ${synchronization_sources})
 endforeach(src)
 
 
-# exactly one of these must be enabled. below is some code to ensure this.
-option(ARGO_VM_SHM
-	"Handle virtual addresses using POSIX shared memory. Size-limited." ON)
-option(ARGO_VM_ANONYMOUS
-	"Handle virtual addresses using anonymously-mapped memory. Slow." OFF)
-option(ARGO_VM_MEMFD
-	"Handle virtual addresses using an anonymous memory file. Requires kernel 3.17+." OFF)
-
-set(ARGO_VM_COUNT 0 CACHE INTERNAL "Check for multiple virtual address handlers")
 unset(vm_libs)
 
-if(ARGO_VM_SHM)
+# Exactly one of these must be enabled (or use the default)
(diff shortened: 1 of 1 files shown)


davidklaftenegger:
No strong objection.

untested:
```
set(ARGO_VM_OPTIONS ANONYMOUS MEMFD SHM )
set(ARGO_VM SHM CACHE STRING "Select how to handle virtual addresses")
set_property(CACHE ARGO_VM PROPERTY STRINGS ${ARGO_VM_OPTIONS})
if(NOT ARGO_VM IN_LIST ARGO_VM_OPTIONS)
    message(FATAL_ERROR "ARGO_VM must be one of ${ARGO_VM_OPTIONS}")
endif()
```

something like this could be used to restrict the string to valid/expected values.
Do we want that, or is it okay that users have to type the string without typo in the command line or their cmake UI of choice to get the behaviour they want?

kostis:
> No strong objection.

What does this comment mean?  That you only have a _weak_ objection?  If so, it is better to explicitly state it than having us guess what this objection might be.


Note that the code in the PR checks for *two* of the values (currently `MEMFD` and `ANONYMOUS`) and if it is not one of them the third value (`SHM`, which is the current default) is used, even if the user has typed something else.  This allows cmake _without_ having to specify something for `ARGO_VM` and get the default configuration, which is the current behaviour.


Regarding:
> something like this could be used to restrict the string to valid/expected values. Do we want that, or is it okay that users have to type the string without typo in the command line or their cmake UI of choice to get the behaviour they want?

IMO, this is an orthogonal issue and something that we can discuss more generally than _just_ in the context of selecting an `ARGO_VM` model.  Note that we can currently happily build the system after a cmake of the form:
```sh
$ cmake -DARGO_BACKEND_MPI=YES_PLEASE   -DBUILD_DOCUMENTATION=ALL_OF_IT ../
```
Why should the `ARGO_VM` option be special?

lundgren87:
I agree that it would be good to either:

- Restrict the possible input space of the ArgoDSM CMake options AND/OR
- Inform when an invalid option is given and what the outcome is (error / use default)

This concerns all options and could be a separate PR.

lundgren87:
> No strong objection.
> 
> untested:
> 
> ```
> set(ARGO_VM_OPTIONS ANONYMOUS MEMFD SHM )
> set(ARGO_VM SHM CACHE STRING "Select how to handle virtual addresses")
> set_property(CACHE ARGO_VM PROPERTY STRINGS ${ARGO_VM_OPTIONS})
> if(NOT ARGO_VM IN_LIST ARGO_VM_OPTIONS)
>     message(FATAL_ERROR "ARGO_VM must be one of ${ARGO_VM_OPTIONS}")
> endif()
> ```
> 
> something like this could be used to restrict the string to valid/expected values. Do we want that, or is it okay that users have to type the string without typo in the command line or their cmake UI of choice to get the behaviour they want?

I think the first three lines of David's suggestion is an elegant solution regardless of validating input (following three lines), as this will present itself as a multiple choice when using `ccmake` letting the user toggle between the three valid options only.

kostis:
> I think the first three lines of David's suggestion is an elegant solution regardless of validating input (following three lines), as this will present itself as a multiple choice when using ccmake letting the user toggle between the three valid options only.

Adopted.  But two lines suffice for this.

lundgren87 (approved):
Looks fine to me.

PR #119, opened 2022-06-17
Substitute C-style casting with C++ one

kostis:
While at it, fix some whitespace issues around the lines that have been
changed.


lundgren87 (dismissed):
The only comment I have is that our code style states that pointers and references should attach to the type, which is not the case on L813 and L1192-1193 in `swdsm.cpp`.

This might be out of the scope of this PR and is violated in _many_ places with old code, so it can also be left for later.

PR #120, opened 2022-06-17
Do not indent inside namespaces

kostis:
There is really no need to indent code inside namespaces.

While at it, cleaned up various places in the code (mostly in comments that now fit in single lines).

Before we merge this, we should discuss whether to adopt the C++17 way of writing nested namespaces.

### DIFF
--- a/src/synchronization/intranode/mcs_lock.hpp
+++ b/src/synchronization/intranode/mcs_lock.hpp
@@ -24,7 +24,7 @@ namespace locallock {
  * @warning The MCS need to be locked and unlocked by the same thread.
  */
 class mcs_lock {
-private:
+ private:
 	/** @brief Thread nodes for the lock */
 	struct mcs_node {
 		/** @brief Construct a new node */
@@ -42,7 +42,7 @@ private:
 	/** @brief Local nodes for the different locks each thread can have */
 	thread_local static std::map<mcs_lock*, mcs_node> selfs;
 
-public:
+ public:
 	/**
 	 * @brief Construct an MCS lock
 	 */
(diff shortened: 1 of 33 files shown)


src/synchronization/intranode/mcs_lock.hpp
```diff
@@ -24,7 +24,7 @@ namespace locallock {
  * @warning The MCS need to be locked and unlocked by the same thread.
  */
 class mcs_lock {
-private:
+ private:
```
lundgren87, line 27: The private/public(below) indentation is a tab in most other places, which is what I would normally expect.

lundgren87 (changes requested):
One comment left above.

Overall I have no issues with the change of code style but I think that this should be accompanied with a change to `docs/code-style.md`, as this currently states:

> Spacing
> 2. Namespaces, as well as public/private/protected labels should be indented

Namespaces should either simply be removed (if indentation of namespaces is "optional") or added as a third bullet stating that namespaces should not be indented. Optionally, if more namespace and/or code-style related changes are planned this could merit a separate PR.

src/synchronization/intranode/mcs_lock.hpp
kostis: Let me undo this - as it does not belong in this PR

lundgren87 (approved):
No more objections from my side. Not sure if anyone else has something to add but I will approve.

kostis:
Any opinions on whether to adopt the C++17 way of writing nested namespaces as part of this PR or not?

lundgren87:
> Any opinions on whether to adopt the C++17 way of writing nested namespaces as part of this PR or not?

I think this is a readability improvement but that it does not necessarily have to be part of this PR. No objection if you want to include it.

kostis:
Let me merge this in order to not unnecessarily delay two other, more important, PRs that I want to submit.

PR #122, opened 2022-06-23
Place include .hpp headers alphabetically

kostis:
Their proper ordering was missed when the tests files were cleaned up.

### DIFF
--- a/tests/allocators.cpp
+++ b/tests/allocators.cpp
@@ -11,10 +11,10 @@
 #include <iostream>
 #include <tuple>
 // ArgoDSM headers
-#include "argo.hpp"
 #include "allocators/collective_allocator.hpp"
 #include "allocators/generic_allocator.hpp"
 #include "allocators/null_lock.hpp"
+#include "argo.hpp"
 #include "backend/backend.hpp"
 // GoogleTest headers
 #include "gtest/gtest.h"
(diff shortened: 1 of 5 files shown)


ioanev:
> **Note**
The top-level ArgoDSM header was purposely left to precede the rest of the headers in that group.

lundgren87 (approved):
Alphabetical order makes sense in order to eventually automate checking.

kostis:
If there are reasons to include header files in some specific order, then we should either find some alternative (e.g. include `argo.hpp` in the other header files that need it) or, of these reasons are very good ones, explicitly document them (perhaps only in some comment) and add a LINT exception.

Leaving such _intentions_ undocumented is not good for understandability and maintenance of the code.  In this particular case, there does not appear to be any good reason to not stick to alphabetic order.  If @ioanev knows such a a reason, perhaps we can document it.

ioanev:
Wrong judgement from my side, and I should have mentioned this as a note in the relevant PR's description.
This was easy miss in the review.
Please go ahead.

> **Note**
The change might also need to be applied here:

https://github.com/etascale/argodsm/blob/2602a69da6f682696a4e012cd91d614671669ac0/src/argo.cpp#L7-L14

kostis:
> > Note
> > The change might also need to be applied here:
> 
> https://github.com/etascale/argodsm/blob/2602a69da6f682696a4e012cd91d614671669ac0/src/argo.cpp#L7-L14

This is a special case.  A `.cpp` file should _always_ include its corresponding `.hpp` (`.h`) file first.

PR #123, opened 2022-06-23
Add ARGO_VM to actions matrix

kostis:
This allows us to test _two_ VM configuations.
The matrix should also be extended with MEMFD when we make this configuration work on "regular" Linux machines.


lundgren87 (approved):
Looks fine, opened #124 regarding investigating MEMFD segfaulting.

PR #129, opened 2022-06-23
Make function `self_upgrade()` local to `swdsm.cpp`

kostis:
This function is only used in the `swdsm` file, so there is no need to
expose it to the rest of the world.  Declare it local and remove its
declaration from the `swdsm.hpp` header file.

### DIFF
--- a/src/backend/mpi/swdsm.cpp
+++ b/src/backend/mpi/swdsm.cpp
@@ -20,8 +20,6 @@ namespace vm = argo::virtual_memory;
 namespace sig = argo::signal;
 namespace env = argo::env;
 
-using namespace argo::backend;
-
 /*Barrier*/
 /** @brief  Locks access to part that does SD in the global barrier */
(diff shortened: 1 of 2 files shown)


src/backend/mpi/swdsm.cpp
```diff
@@ -1012,7 +1016,13 @@ void self_upgrade(upgrade_type upgrade) {
...
 }
 
-void swdsm_argo_barrier(int n, upgrade_type upgrade) {
+/**
+ * @brief Global barrier for ArgoDSM - needs to be called by every thread in the
+ *        system that need coherent view of the memory
+ * @param n number of local thread participating
+ * @param upgrade the type of classification upgrade to perform
+ */
+void swdsm_argo_barrier(int n, argo::backend::upgrade_type upgrade) {
```
lundgren87, line 1025: This now contains the same documentation in two places. Is it not simpler to keep this to the header so that future changes do not have to be replicated in two files?

lundgren87 (approved):

PR #131, opened 2022-06-29
Install ArgoDSM headers directly under include

lundgren87:
The ArgoDSM headers are currently installed to
`ARGO_INSTALL_DIRECTORY/include/argo`, which is both
confusing as this breaks absolute include paths internally and
requires external include paths to specify the argo prefix.
Installing the ArgoDSM headers to `ARGO_INSTALL_DIRECTORY/include`
directly while maintaining the directory structure resolves this.

### DIFF
--- a/docs/_includes/argo_example.cpp
+++ b/docs/_includes/argo_example.cpp
@@ -5,7 +5,7 @@
 
 #include <pthread.h>
 
-#include "argo/argo.hpp"
+#include "argo.hpp"
 
 struct thread_args {
 	int data_begin;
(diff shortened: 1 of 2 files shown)


lundgren87:
The side-effect of this change is that include paths may have to be updated in external applications.

The correct way to specify include path is now (as I believe you can expect):
`-I${ARGO_INSTALL_DIRECTORY}/include`
And the correct way to include `argo.hpp` in external applications is now:
`#include "argo.hpp"` or `#include "argo.h"`

kostis (approved):
Looks good , of course.

In the interest of achieving progress fast, I will approve it, but ideally I would have liked to also include a suitable test that checks that we can build external applications when installing in the "new way" and following the suggested way to include `argo.hpp`.

lundgren87:
On second thought, this may need rethinking as it would clutter the default include path with a number of directories that on their own make no sense (eg `/usr/local/include/backend/mpi`).

PR #133, opened 2022-11-23
Make sure the C interface only uses C stuff

kostis:
The C header file came with some C++ stuff (`argo::node_id_t`) and also used `int` instead of `unsigned int` as the return type of some functions returning node_id's. This PR makes all these `unsigned ints` instead.


lundgren87 (approved):
These look like reasonable changes.

PR #134, opened 2022-11-23
Test argo_example.cpp in the Actions workflow

lundgren87:
This PR introduces new steps to the GitHub actions workflow that build and test [argo_example.cpp](docs/_include/argo_example.cpp). This requires ArgoDSM to be installed, so this step is also added. I believe that this will expose errors related to building, linking and running external applications that use ArgoDSM that we have previously had to rely on running Jenkins separately in order to catch.

Note that the ArgoDSM installation uses roughly 40Mb storage, but I have not found any restrictions on GitHub actions storage for public repositories.

_NOTE_: This PR passes `-DPAGE_SIZE=4096UL` as a parameter to mpic++ in order to successfully complete the GitHub Actions workflow. This is a bug and should be logged and fixed separately.


kostis (changes requested):
I think that for the time being we should also modify the `tutorial.md` to also mention the `-DPAGE_SIZE=4096UL` argument in the `mpic++` call.  The rationale is that the tutorial should reflect whatever a user needs to do to compile and run a program in current Argo, even if that is to change in the (near) future.

lundgren87:
> I think that for the time being we should also modify the `tutorial.md` to also mention the `-DPAGE_SIZE=4096UL` argument in the `mpic++` call. The rationale is that the tutorial should reflect whatever a user needs to do to compile and run a program in current Argo, even if that is to change in the (near) future.

I suggested a permanent solution to the issue in #137 and I think that should perhaps be examined before introducing a temporary workaround to the documentation.

kostis:
My comment was more related to the fact that this PR is a bit inconsistent: it leaves the `mpic++` call in the tutorial unchanged (i.e. without mentioning the need for using a `-DPAGE_SIZE` argument), but at the same time _uses_ such an argument in the `mpic++` call in the `make_external.sh` script.

We can certainly merge #137 before this one and then there is no need for changes in the `mpic++` call in the tutorial.  But we should then take out the `-DPAGE_SIZE` from the `make_external.sh` script.

kostis (approved):

lundgren87:
> My comment was more related to the fact that this PR is a bit inconsistent: it leaves the `mpic++` call in the tutorial unchanged (i.e. without mentioning the need for using a `-DPAGE_SIZE` argument), but at the same time _uses_ such an argument in the `mpic++` call in the `make_external.sh` script.
> 
> We can certainly merge #137 before this one and then there is no need for changes in the `mpic++` call in the tutorial. But we should then take out the `-DPAGE_SIZE` from the `make_external.sh` script.

Just for clarity, `-DPAGE_SIZE=4096UL` was removed from the `make_external.sh` script as it is redunant following the merge of #137 .

PR #136, opened 2022-11-23
Remove unnecessary .hpp includes from argo.hpp

kostis:
In contrast to argo.h, the argo.hpp file contained many .hpp include files. Both compilation and running ArgoDSM seem to work OK when these includes are not present, so this PR simply removes them.

### DIFF
--- a/src/argo.hpp
+++ b/src/argo.hpp
@@ -10,10 +10,7 @@
 #include <cstddef>
 
 #include "allocators/allocators.hpp"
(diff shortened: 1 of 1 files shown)


lundgren87 (approved):
It does not appear that this breaks any functionality in the main ArgoDSM API.

I believe it is reasonable to ask for manual inclusion of `backend/backend.hpp` should you wish to access backend functionality directly.

PR #139, opened 2022-11-28
Add a cpplint GitHub action

kostis:
Add a test for code style checking, using cpplint. Checks which are not (yet) satisfied by the ArgoDSM code base are explicitly disabled.  Perhaps some of them are a good idea to enable.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
       ARGO_VM:  ${{ matrix.vm  }}
     steps:
       - name: Checkout code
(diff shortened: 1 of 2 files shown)


lundgren87 (approved):
I suggest we use `actions/checkout@v3`, I will leave it up to you if you want to do this in a commit before merging or in a separate PR for all actions, it is not urgent. For reference see:
https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

lundgren87 (approved):
Looks good!

PR #140, opened 2022-11-29
Name constants with a leading "k" and mixed case

kostis:
A C++ programming convention, also suggested by the [Google style guide](https://google.github.io/styleguide/cppguide.html#Constant_Names) and enforced by `cpplint`, is to name variables declared `constexpr` or `const`, and whose value is fixed for the duration of the program, by a leading "k" followed by mixed case.

There are only two files that require changes, one in the main source, which may require some discussion, and one in the test suite, which is uncontroversial, I think.  This commit fixes both of them.

### DIFF
--- a/tests/lock.cpp
+++ b/tests/lock.cpp
@@ -30,7 +30,7 @@ constexpr std::size_t size = 1<<20;
 constexpr std::size_t cache_size = size;
 
 /** @brief number of threads to spawn for some of the tests */
-constexpr int nThreads = 16;
+constexpr int kThreads = 16;
 /** @brief number of itereations to run for some of the tests */
 constexpr int iter = 10000;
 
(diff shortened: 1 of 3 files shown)


tests/lock.cpp
```diff
@@ -30,7 +30,7 @@ constexpr std::size_t size = 1<<20;
 constexpr std::size_t cache_size = size;
 
 /** @brief number of threads to spawn for some of the tests */
-constexpr int nThreads = 16;
+constexpr int kThreads = 16;
 /** @brief number of itereations to run for some of the tests */
 constexpr int iter = 10000;
```
lundgren87, line 35: I fail to understand what the difference between L33 and L35 is?
It seems to me that if one adopts the `kMixedCase` convention, should this not apply to both?

lundgren87 (commented):
I've mentioned one case inline as an example, but in general I am not sure why this should not apply to all static/global `const`-declared variables in that case?

I think there are quite a few more that violate this code standard should we adopt it (which is perhaps the real question?).

kostis:
My guess is that the `cppcheck` rule only applies to variables either containing underscores or already in CamelCase but leaves _simple_ variables (e.g. `i`, `n`, `iter`, ...) without complaints.

lundgren87:
Does the change from `page_info_size`->`kPageInfoSize` perhaps conflict with our code style guide?

1. All namespaces, classes, class members, functions, and variables should be in snake_case.
2. “Magic” constants, as well as C-style enumerations that are used as integers should be CAPITALIZED.

tests/lock.cpp
kostis: You have a point there.

My guess is that the `cpplint` check(s) concern only variables that are either in camel case (e.g. `nThreads`) or in snake case (e.g. `page_info_size`), and leave variables such as `iter` which are neither unchanged.

PR #141, opened 2022-11-30
Remove unneeded `exit` calls after `throw` lines

### LINKED ISSUE #97: Call to `exit()` after `throw`
In many places in the code, I see the following pattern:
```c++
   ...
   throw std::system_error(std::make_error_code(static_cast<std::errc>(errno)), ...);
   exit(EXIT_FAILURE);
```
What _exactly_ is the purpose of writing this?  Don't we trust that the `throw` call will throw an exception?

Any suggestions on how to re-write such code?

kostis:
It is not a good idea to have dead (and potentially confusing) code in the code base; esp. if this is in code which is not tested in any way in ArgoDSM's testsuite.  This PR removes it.

Closes #97.


lundgren87 (approved):
For the sake of consistency I suggest we approve this. We currently do not program defensively against disabled exception handling in general (env, mempools and backend modules).

PR #142, opened 2022-12-08
Move VM reservation to fix MEMFD error

### LINKED ISSUE #124: Using the MEMFD VM option segfaults on up-to-date Debian and Ubuntu
This is not resolved by setting `vm.overcommit_memory` to 1 and should be investigated.

This can be reproduced on GitHub actions using `ubuntu-latest`, as well as on Ubuntu 22.04 and up-to-date Debian.

lundgren87:
This PR fixes #124 by moving the start of ArgoDSM's Virtual Memory reservation from `0x200000000000` to `0x155555554000`. The start of the reservation is moved from 1/4 into the user-space VM to 1/6 into the user-space VM.

The reason for this change is that for x86_64 specifically, the area beginning at `0x555555554000` (2/3 into the user-space VM) is reserved for PIE loads followed by the heap, and ArgoDSM's current VM reservation intrudes on this area for MEMFD leading to segmentation faults during ArgoDSM initialization. See below for reference:
https://github.com/torvalds/linux/blob/ce19275f0103934828cb19712b6d8552c39476c8/arch/x86/include/asm/elf.h#L237-L243

The same change is applied for SHM and ANONYMOUS for the sake of consistency. An alternative change could be to move this allocation to `0x100000000000` for the sake of easier debugging, but there is functionally no difference.

Finally, MEMFD is added to the GitHub Actions matrix. To facilitate this, `vm.overcommit_memory` is set to 1 (`always overcommit`) ahead of `make test` as this is a requirement for MEMFD.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         dbg: [OFF, ON]
-        vm: [ANONYMOUS, SHM]
+        vm: [ANONYMOUS, SHM, MEMFD]
     name: On ${{ matrix.os }} (DBG=${{ matrix.dbg }}, VM=${{ matrix.vm }})
     runs-on: ${{ matrix.os }}
     env:
@@ -29,6 +29,8 @@ jobs:
       - name: Build external
         run: ./.github/scripts/make_external.sh
       - name: Run test suite
-        run: make test -C build
+        run: |
+          sudo sysctl -w vm.overcommit_memory=1
+          make test -C build
       - name: Run external
         run: ./.github/scripts/run_external.sh
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         dbg: [OFF, ON]
-        vm: [ANONYMOUS, SHM, MEMFD]
+        vm: [ANONYMOUS, MEMFD, SHM]
     name: On ${{ matrix.os }} (DBG=${{ matrix.dbg }}, VM=${{ matrix.vm }})
     runs-on: ${{ matrix.os }}
     env:
--- a/src/virtual_memory/anonymous.cpp
+++ b/src/virtual_memory/anonymous.cpp
@@ -24,9 +24,22 @@
 
 namespace {
 /* file constants */
-/** @todo hardcoded start address */
-char* const ARGO_START = reinterpret_cast<char*>(0x200000000000l);
-/** @todo hardcoded size */
+/**
+ * @brief The start of the ArgoDSM virtual memory space
+ * @note This hard-coded value assumes x86_64 architecture
+ *
+ * The ArgoDSM virtual memory space leaves the first 1/6 for local use.
+ */
+char* const ARGO_START = reinterpret_cast<char*>(0x155555554000l);
+/**
+ * @brief The size of the ArgoDSM virtual memory space
+ * @note This hard-coded value assumes x86_64 architecture
+ *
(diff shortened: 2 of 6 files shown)


lundgren87:
No regression on Jenkins with `MEMFD` enabled. @davidklaftenegger please feel free to correct me if I misunderstood or left something out following our discussion regarding this.

davidklaftenegger:
Your explanation matches our discussions on the topic. However, I believe the size in SHM and ANONYMOUS also should be adjusted to no longer be able to extend into the range used by dynamically loaded binaries.

davidklaftenegger:
just for clarification, this patch would then make all VM options use up to the range from 0x155555554000 to 0x555555553999 for ArgoDSM. Also on 56bit-pointer architectures (ice lake / zen3) we could use (much) larger ranges, but for now, 64TB maximum memory sounds reasonable.

lundgren87:
> Your explanation matches our discussions on the topic. However, I believe the size in SHM and ANONYMOUS also should be adjusted to no longer be able to extend into the range used by dynamically loaded binaries.

I was confused by this at first, then I realized that `0x80000000000l` is actually 8TB and not 64TB, so the maximum SHM/ANONYMOUS reserved area is significantly smaller and they do not risk extending in to the reserved area. I don't know if there is a reason for this difference in size, hence I left it unchanged.

```
MEMFD: const ptrdiff_t ARGO_SIZE = 0x400000000000l;
SHM  : const ptrdiff_t ARGO_SIZE = 0x80000000000l;
```

Note that I have not tested reserving up to 64TB instead of 8TB for SHM/ANONYMOUS. I don't see it being a problem and can test this change if there's no reason to keep the current smaller sizes.

kostis (commented):
Looks OK overall, but I do not really like this (unnecessary, I think) comment and "hard-coded magic number" code duplication in three files.  Admittedly, some of the duplication was there before, but now that the comments have been extended the issue has become more pronounced.  In any case, can't we factor out all these repeated lines to some common place? (e.g. a new header file)

Also, to a reader, it seems strange that two of the magic SIZE constants are called `ARGO_SIZE` (and their values differ!), but the third one is called `ARGO_SIZE_LIMIT` (and its value is identical to one of the `ARGO_SIZE` anonymous magic constants).  Perhaps we should extend the relevant comments to explain why these values and constant names are chosen.

davidklaftenegger:
I think now that we actually know why these numbers are set as they are (as opposed to "it just works" before), a common header makes sense.

.github/workflows/ci.yml
```diff
@@ -9,7 +9,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         dbg: [OFF, ON]
-        vm: [ANONYMOUS, SHM]
+        vm: [ANONYMOUS, SHM, MEMFD]
```
kostis, line 12: Really minor: perhaps `MEMFD` should be placed before `SHM` to preserve alphabetical order.

davidklaftenegger:
> ```
> MEMFD: const ptrdiff_t ARGO_SIZE = 0x400000000000l;
> SHM  : const ptrdiff_t ARGO_SIZE = 0x80000000000l;
> ```

oops. well, I see no reason to keep them different.

> Note that I have not tested reserving up to 64TB instead of 8TB for SHM/ANONYMOUS. I don't see it being a problem and can test this change if there's no reason to keep the current smaller sizes.

Please do.

lundgren87:
I'll test the functional changes tomorrow and address the remaining comments after this.

Regarding putting the sizes in a header, we have `virtual_memory.hpp` but this is a nice, clean file with no implementation details. Perhaps a separate header containig the sizes would be better?

kostis (approved):

kostis (approved):

PR #146, opened 2023-10-16
Initialize barrierlockholder to prevent maybe-uninitialized

suyashmahar:
Current master branch fails to compile with gcc 9.4:

```
src/backend/mpi/swdsm.cpp:1058:2: error: ‘barrierlockholder’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
  if (pthread_equal(barrierlockholder, pthread_self())) {
  ^~
cc1plus: all warnings being treated as errors
```


davidklaftenegger:
Thank you for bringing this issue to our attention.
Your patch would silence the warning, but there is an underlying potential problem in that the code may still misbehave if `pthread_self() == pthread_t()`, which seems to be permitted by POSIX.

PR #147, opened 2023-10-19
Fix 2 uninitialized variables

kostis:
Perform a code rewrite and a variable initialization to shut off (newer) compiler warnings about unitialized variables.

Supersedes #146


davidklaftenegger (approved):
Now that I think about it, another option would have been to declare the `barrierlockholder` static (and enforce correct memory ordering), but I like this solution much better.

PR #153, opened 2025-09-12
Upgrade to Doxygen 1.9.8 (from 1.9.1)

kostis:
This shuts off many deprecation warnings on newer doxygen versions.


lundgren87 (approved):
I don't see any issues with this. We have a lot of undocumented parameters in the tests but that is a separate issue.

